PR #78, opened 2026-05-31
fix(benchmarks): Native eig uses generic solver; add GFLOP/s-vs-N plotting

Ravenwater:
Two related benchmark improvements (separate commits).

## 1. Fix the eig Native-bypass gap

`eigenvalue_sym_op<Native>` called `mtl::eigenvalue_symmetric(A)`, which
auto-dispatches to **LAPACK** `syev` when `MTL5_HAS_LAPACK` is set — so the
`native` and `lapack` benchmark rows were identical (~1.0× speedup) and the true
cost of the C++ eigensolver was hidden.

- Extracted the generic tridiagonalize + implicit-QR algorithm out of
  `eigenvalue_symmetric` into a public **`eigenvalue_symmetric_generic()`**. The
  dispatcher calls it on the non-LAPACK path — public API behaviour unchanged.
- Pointed the Native backend at `eigenvalue_symmetric_generic()` so it never
  reaches LAPACK.

Now the Native row reflects the reference path — e.g. at n=256 it is **~10.8×
slower** than the LAPACK/MKL path (was reported as 1.0×).

**Validation:** eigen unit tests pass with gcc both *without* LAPACK (generic
path) and *with* LAPACK (dispatch path); clang `-Wall -Wextra` clean.

## 2. GFLOP/s-vs-N plotting + example data

- **`benchmarks/plot_results.py`** — matplotlib tool that plots bench_all CSV
  output as GFLOP/s (or time) vs N. Accepts multiple CSVs and overlays them, so
  native/OpenBLAS/MKL renders in one figure. Stdlib-only besides matplotlib;
  headless-safe (`Agg`); `--op`, `--metric`, `--labels`, `--logx/--logy`.
- **`benchmarks/data/`** — small committed example sweeps (odd-size BLAS sweep
  `65:1025:80`, single-threaded; OpenBLAS + MKL) so the curves reproduce without
  re-running the suite.
- README documents both.

Example three-way overlay (`--op gemm`): MKL ≈ 71–74 GFLOP/s edges OpenBLAS, and
the two `native` curves coincide at ~5–8 GFLOP/s — visually confirming fix #1
(both `native` series are the same C++ code).

> Note: this Python is benchmark *tooling*, not library bindings — the
> NumPy/SciPy bindings remain in the separate `mtl5-python` repo.

## Changes
- `include/mtl/operation/eigenvalue_symmetric.hpp` — extract `eigenvalue_symmetric_generic`.
- `benchmarks/harness/op_lapack.hpp` — Native eig → generic.
- `benchmarks/plot_results.py` (new), `benchmarks/data/*.csv` (new), `benchmarks/README.md`.

## Test plan
- [ ] Fast CI passes (the eig refactor is the CI-relevant part; benchmarks aren't built in CI)
- [ ] Promote to ready: `gh pr ready <N>`

## Summary by CodeRabbit

* **Documentation**
  * Expanded benchmark docs with a new "Plotting" guide, examples, and committed reference CSVs for reproducible curves.

* **New Features**
  * Added a standalone Python CLI to plot benchmark CSVs with selectable operations, metrics, labels, log scaling, and PNG export.

* **Refactor**
  * Reworked symmetric eigenvalue path to provide a LAPACK-aware dispatcher with a LAPACK-free generic fallback for improved modularity.


benchmarks/data/blas_sweep_mkl.csv
```diff
@@ -0,0 +1,105 @@
+operation,backend,size,median_ns,min_ns,max_ns,mean_ns,stddev_ns,gflops,iterations
```
Ravenwater, line 1: we do not have any indication of the platform on which this was run

Ravenwater (commented):
We cannot have raw CSV files without any attribution to the platform on which they were gathered. Either annotate the files, or create a data directory with the platform label and provide details on the hardware configuration that yielded the data.

benchmarks/data/blas_sweep_mkl.csv
Ravenwater: Good catch — added `benchmarks/data/README.md` (164a6e6) recording the platform: i7-12700K, Ubuntu 24.04, GCC 13 -O3, OpenBLAS 0.3.26 / MKL oneAPI 2026.0, single-threaded, plus the exact generation commands and a 'reference numbers, not a claim' caveat.

benchmarks/plot_results.py
```diff
@@ -0,0 +1,148 @@
...
+    # Blank any unused subplots.
+    for j in range(len(ops), nrows * ncols):
+        axes[j // ncols][j % ncols].axis("off")
+
+    fig.suptitle(f"MTL5 benchmark: {ylabel} vs N", fontsize=13)
+    fig.tight_layout(rect=(0, 0, 1, 0.97))
+
+    if args.show:
+        matplotlib.use("TkAgg", force=True)  # best-effort interactive
+        plt.show()
```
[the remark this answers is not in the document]
Ravenwater: Fixed in 164a6e6 — args are parsed first and the backend is chosen before pyplot is imported (Agg only when not --show), so --show now works and the no-op late use("TkAgg") is gone.

Issue #186, opened 2026-06-23
Epic: native supernodal LU (SuperLU) parity with SuiteSparse SuperLU


Ravenwater:
## Goal

Build a **native supernodal unsymmetric LU** and make it a genuine alternative to the external SuiteSparse SuperLU binding for general unsymmetric matrices — especially the fill- and flop-heavy ones where supernodal dense panels pay off. SuiteSparse SuperLU is a proof of existence: the performance it delivers is the **bare minimum** native SuperLU must match. The *ceiling* is higher than SuperLU can reach — a native, number-type-generic implementation gains **mixed precision** (low-precision supernode panels + high-precision accumulation + iterative refinement, with custom types like posit/LNS), which the fixed-precision C library cannot offer.

Plan: [`docs/plans/native-superlu-performance.md`](../blob/main/docs/plans/native-superlu-performance.md).

## Build-then-tune — the key difference from #138

#138 (native KLU) was a *parity* effort: native KLU already existed and competes in the **same algorithm class** as SuiteSparse KLU (scalar, non-supernodal Gilbert–Peierls LU), so its gap was pure constant-factor.

SuperLU is a **different algorithm class**: supernodal left-looking LU with **dense BLAS-3 panel updates** and **threshold partial pivoting**. Today's native LU is scalar and column-at-a-time, so closing the gap requires *building* the supernodal machinery for the unsymmetric, pivoting case — not just removing overhead.

Most of that machinery now exists and is reused:
- **#178** — native supernodal symmetric LDLᵀ: supernode detection, the dense-panel left-looking kernel, and the `accumulator_traits` boundary.
- Mixed-precision dense kernels: widening dot/GEMV, accumulator policy, custom-accumulator dispatch (#160, #162, #163, #164, #165).
- Generic iterative refinement (#119) and the SuperLU/UMFPACK/CHOLMOD bindings as oracles.

The unsymmetric case adds the **column elimination tree**, **unsymmetric supernode symbolic**, and the hard part — **panel partial pivoting interleaved with supernodal updates**. The dense block update reuses the existing mixed-precision GEMM path.

## Baseline gap

[...]

Issue #183, opened 2026-06-23
perf(sparse): 2D panel blocking + supernode-panel update efficiency


Ravenwater:
Phase 3 of native supernodal LU (SuperLU) performance parity (see docs/plans/native-superlu-performance.md). Part of the native SuperLU performance epic.

## Work
- 2D blocking of the supernode-panel update; cache / BLAS-3 tuning of the dense kernels.
- Tighten gather/scatter and relative-index handling in the panel update.

## Acceptance
- Native within ~1.5-2x of SuperLU factor+solve time across the Phase 0 suite.

Ravenwater:
## Performance parity is a dedicated future project (status update)

Phases 3a/3b/3c explored making native `supernodal_lu` competitive with SuiteSparse SuperLU. **Conclusion from three measured iterations: incremental work on the current data structure does not reach parity, and that pursuit is parked here as a dedicated future project.**

### What was tried (all correct, all measured)
- **3a — BLAS-2 block update** (per-column TRSV+GEMV via `accumulator_traits`/BLAS): ~scalar parity; fixed the Phase-2 regression but no net win.
- **3b — panel BLAS-3** (batched cross-supernode TRSM+GEMM over W-column panels): PR #192 (closed unmerged). Correct, ~scalar parity. Bottleneck diagnosed: indirect gather/scatter to a flat `n×W` workspace + small (width-16) GEMMs.
- **3c — subset tracking** (apply each supernode only to reached columns): correct; still parity. Widening panels (16→64) didn't help.

### Measured (BLAS, single-thread)
| matrix | panel (3c) | scalar `sparse_lu` | SuperLU |
|---|--:|--:|--:|
| dense_400 | 33 ms | 33 ms | 8 ms |
| dense_1200 | 0.50 s | 0.54 s (0.94×) | — |
| banded_5k | 46 ms | 37 ms (1.26×) | 25 ms |

The native solver only reaches parity-with-scalar at large dense (0.94×) and is ~1.3× slower on banded; the scalar baseline is itself **~4× off SuperLU**.

### What parity actually requires (the future project)
Closing the ~4× gap to a 25-year-tuned C library needs the **full supernodal data-structure overhaul together**, not incrementally:
- supernodal-**aligned storage** (contiguous supernode rows; no hot-loop indirection),
- **larger supernodes + 2-D blocking** (so GEMM K-dimension is large enough for peak BLAS),
- a rewritten **panel symbolic** (eliminate the current double-reach),
[...]

PR #192, opened 2026-06-25
perf(sparse): panel BLAS-3 cross-supernode update for supernodal LU (#183)

Ravenwater:
## Summary

**Phase 3b** of the native-SuperLU epic (#186). Restructures the supernodal LU numeric loop into **panels** (panelW=16) with supernodes aligned to panel boundaries, so each earlier (closed) supernode applies to a whole panel as one **TRSM + GEMM (BLAS-3)** — routed through `interface::blas` (or native-SIMD `gemm_blocked`, or a scalar fallback) when `Accumulator==Value` and the value type is float/double. Custom accumulators keep the proven Phase-2 per-column path.

This is landed as **correct groundwork**, not yet a speedup — see the honest status below.

## How it works

An `n × panelW` dense workspace. At each panel start the straddling supernode is closed (so every column `< kp` is in a *closed* supernode), and every closed supernode contributing to the panel is applied via **batched GEMM** across all panel columns at once. Intra-panel columns then update each other with the scalar SAXPY and are pivoted / emitted exactly as Phase 2. A naturally-wide supernode split across panels gets its trailing updates as cross-panel GEMM too.

The key correctness subtlety (and the bug I had to fix): the panel-boundary close must set `Lp[kp]` *before* building the closing supernode's dense block, or the last column's entries are silently dropped — only visible once a supernode is used as a batch source (n > panelW).

## Correctness

Validated **bit-for-result against scalar `sparse_lu` to machine precision** across dense / banded / 2-D convection-diffusion / random and multiple panels (dedicated multi-panel test added). All build configs (no-BLAS / `MTL5_WITH_BLAS` / native-fast) green; warning-clean GCC + Clang. 358 assertions in the supernodal-LU suite; full sparse suite 28/28.

## Honest performance status — groundwork, not a win yet

| | Phase-2 (merged) | this (panel) | scalar `sparse_lu` | SuperLU |
|---|--:|--:|--:|--:|
| dense_200 | 16.6 ms | 6.5 ms | 6.2 ms | 1.9 ms |
| banded_5k | 42 ms | 38 ms | 35 ms | 25 ms |

This **fixes the Phase-2 regression** (the supernodal solver was *slower* than scalar; now ~parity) and does genuine BLAS-3, but lands at **~scalar parity** — it does **not** beat scalar or reach SuperLU parity. Root cause, clearly diagnosed: the batched GEMM is gated by the **indirect gather/scatter** between the flat workspace and the supernode rows (`S.rows`). Widening panels (16→64) didn't help, confirming the bottleneck is indirection, not GEMM size.

**Next step for actual parity (#183 follow-up):** supernodal-**aligned storage** — store L and the workspace so a supernode's rows are contiguous, eliminating hot-loop indirection (SuperLU's actual data structure). That's a further large structural change; this panel restructure is its prerequisite.

I'm landing this per maintainer direction as the correct BLAS-3 panel groundwork that the aligned-storage work builds on. Reasonable to question merging a parity-only change — happy to instead hold it on the branch until the aligned-storage step lands if preferred.

## Summary by CodeRabbit

* **New Features**
  * Improved sparse LU factorization performance by introducing a faster panel-based update path for dense supernodes.
  * Expanded automated validation with a new LU consistency test covering cases that span multiple panel/supernode regions.

* **Bug Fixes**
  * Fixed factorization behavior across panel boundaries to keep results consistent for wide dense supernodes.
  * Improved alignment between the optimized supernodal solver and the scalar reference solution.

### DIFF
--- a/include/mtl/sparse/factorization/supernodal_lu.hpp
+++ b/include/mtl/sparse/factorization/supernodal_lu.hpp
@@ -43,6 +43,11 @@
 #include <mtl/sparse/analysis/column_etree.hpp>
 #include <mtl/sparse/factorization/triangular_solve.hpp>
 #include <mtl/sparse/factorization/sparse_lu.hpp>   // lu_symbolic conventions, accumulator_traits
+#include <mtl/interface/blas.hpp>                    // blas::gemv / blas::trsv (MTL5_HAS_BLAS)
+#include <mtl/interface/dispatch_traits.hpp>         // is_blas_scalar_v
+#ifdef MTL5_NATIVE_FAST_GEMM
+#include <mtl/detail/gemm_blocked.hpp>                // native-SIMD blocked GEMM (panel update)
+#endif
 #include <mtl/sparse/iterative_refine.hpp>
 
 namespace mtl::sparse::factorization {
@@ -161,7 +166,6 @@ supernodal_lu_factor<Value> supernodal_lu_numeric(
 
     using idx32 = std::int32_t;
     std::vector<idx32>          pinv(n, -1);
-    std::vector<Accumulator>    x(n);
     std::vector<idx32>          xi(n);
     std::vector<std::ptrdiff_t> pstack(n);
     std::vector<char>           marked(n, 0);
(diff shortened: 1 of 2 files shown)


Ravenwater:
Closing without merging per maintainer decision. The panel BLAS-3 + subset-tracking work here is **correct** (validated to machine precision, all build configs, dual-compiler clean) but lands at ~scalar parity, not a speedup — see #183 for the measurement-driven conclusion. The branch `perf/superlu-block-kernel` is preserved as groundwork should the full supernodal data-structure overhaul be undertaken later.

PR #272, opened 2026-07-24
feat(generators): range spacing vectors (linspace family) + magic square

Ravenwater:
## Summary
First step of extracting the linear-algebra layer out of Universal into MTL5 (Universal epic stillwater-sc/universal#1204, phase stillwater-sc/universal#1206). The survey of Universal's `sw::blas` vs MTL5 found MTL5 already has essentially the entire BLAS surface -- the only genuinely-missing utilities are the NumPy-style range/spacing vectors and `magic`. This adds them.

## What's added
- **`generators/ranges.hpp`** -- `arange`, `linspace`, `logspace`, `geomspace`, returning `vec::dense_vector<T>`, generic over the scalar type (Universal-free, `mtl::generators` namespace).
  - `arange` -- half-open `[start, stop)` integer-strided (NumPy semantics).
  - `linspace` -- `steps` samples over `[start, stop]`, endpoint pinned exactly; `endpoint=false` for the half-open variant.
  - `logspace` -- `base^linspace(start, stop)` (start/stop are exponents).
  - `geomspace` -- **true geometric progression between the endpoints**. This corrects the historical Universal implementation, which aliased `logspace` and treated the endpoints as exponents (making `geomspace(1,1000,4)` wildly wrong); it now yields `{1, 10, 100, 1000}`.
- **`generators/magic.hpp`** -- magic square of order N: odd via the Siamese method, doubly-even (`N%4==0`) via the complement method. Singly-even orders throw `std::invalid_argument` pending support. Every row/column/diagonal sums to `N(N^2+1)/2`.
- Both added to `generators/generators.hpp`; CHANGELOG updated.

## Tests (`tests/unit/generators/`)
- `test_ranges.cpp` -- 51 assertions / 11 cases (arange half-open + negative step + degenerate; linspace inclusive/exclusive/edge; logspace decades; geomspace decades, powers-of-two, negative same-sign).
- `test_magic.cpp` -- 196 assertions / 6 cases (magic-constant on odd + doubly-even orders, permutation-of-`1..N^2`, known 3x3 values, unsupported-order throws).

| | gcc (ctest) | clang (-Wall -Wpedantic) |
|---|---|---|
| build | OK | OK |
| tests | PASS | PASS (headers) |

## Notes
- Universal-free: no `find_package(universal)`; generic over the `Scalar` type.
- `linspace` was the specifically-requested gap; `geomspace` is delivered corrected rather than a verbatim port of the buggy Universal version.

Part of the Universal->MTL5 BLAS extraction (stillwater-sc/universal#1206).

## Summary by CodeRabbit

- **New Features**
  - Added NumPy-style `arange`, `linspace`, `logspace`, and `geomspace` vector generators with dense output, endpoint control, and input validation (including true geometric progression for `geomspace`).
  - Added `magic` square matrix generator supporting odd and doubly-even sizes, with exceptions for unsupported/invalid orders.
  - Exposed the new generators through the main generator interface.
- **Tests**
  - Added unit tests validating magic-square properties and generator edge cases.
- **Documentation**
  - Updated the unreleased changelog to describe the new generators and semantics.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ Format follows [Conventional Commits](https://www.conventionalcommits.org/).
 
 ### Added
 
+#### Range and test-matrix generators (BLAS extraction from Universal)
+- **`mtl::generators::{arange, linspace, logspace, geomspace}`** (`generators/ranges.hpp`) — NumPy-style spacing vectors returning `vec::dense_vector<T>`, generic over the scalar type (Universal-free). `geomspace` computes a true geometric progression between its endpoints, fixing the historical Universal implementation that aliased `logspace` and treated the endpoints as exponents (`geomspace(1, 1000, 4) == {1, 10, 100, 1000}`).
+- **`mtl::generators::magic`** (`generators/magic.hpp`) — magic square of order N (odd via the Siamese method, doubly-even via the complement method); singly-even orders throw `std::invalid_argument` pending support.
+- Migrated from Universal's `sw::blas` as the first step of extracting the linear-algebra layer into MTL5 (Universal epic #1204, phase #1206).
+
 #### Matrix/vector/tensor property predicate module (#244)
 A cohesive set of runtime property and predicate queries as free functions in `namespace mtl`, built on the existing primitives (cholesky/lu/svd/eigenvalue/norms) with no new dependencies. Consistent tolerance policy: structural checks are exact-by-default and NaN-safe (`!(dev <= tol)`), while norm/factorization/spectral-backed checks use relative or scale-aware tolerances; verified on both the in-house and LAPACK paths.
 - **Structural + vector predicates** (`operation/matrix_properties.hpp`, `operation/vector_properties.hpp`) — `is_square`, `is_empty`, `is_symmetric`, `is_hermitian`, `is_upper/lower/is_triangular`, `is_diagonal`, `is_banded`, `is_diagonally_dominant`; `is_zero`, `is_finite`/`has_nan`/`has_inf`, `is_normalized`/`is_unit`, `is_orthogonal_to`. O(n)/O(nnz), no factorization (#245)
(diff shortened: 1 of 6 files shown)


Ravenwater:
Addressed all 6 review findings in 7d6d256 (verified on GCC 13 and Clang 18).

| # | Finding | Resolution |
|---|---------|-----------|
| 1 | `magic` `N*N` overflow | Guard `N > SIZE_MAX/N` → `std::invalid_argument` before allocation; cache checked `N2` and reuse it in the loops/complements. |
| 2 | `arange` signed overflow (Critical) | Count elements and step magnitude in unsigned arithmetic; skip the unused final `sample += step`. Handles spans near `INT64_MIN`/`INT64_MAX` and `step == INT64_MIN`. Regression test added. |
| 3 | `geomspace` invalid domain | Reject zero and opposite-sign endpoints up front (`std::invalid_argument`) — a progression through zero / across a sign change is undefined. |
| 4 | integral `T` truncates spacing | `static_assert(!std::is_integral_v<T>)` on `linspace`/`logspace`/`geomspace`; custom float-like types (posits) unaffected. |
| 5 | 1×1 magic untested | Added `magic<double>(1)` test (1 row/col, holds 1). |
| 6 | geomspace test coverage | Added interior-value assertions for the negative same-sign case + zero/opposite-sign rejection cases. |

All valid; none skipped.

Ravenwater:
Fixed the follow-up in 33ae8e8. Correct catch — my unsigned count still wrapped for the absolute full-range span (`span + mag - 1` overflows uint64 when `span == UINT64_MAX`, e.g. `arange(INT64_MIN, INT64_MAX, INT64_MAX)` → computed 0 instead of 3). Switched both branches to `span/mag + (span%mag != 0)` (overflow-safe ceiling) and extended the regression test with that full-range case. Verified on GCC 13 and Clang 18.

PR #273, opened 2026-07-24
feat(operation): dense mixed-precision LU iterative refinement + normwise backward error

Ravenwater:
## Summary
Ports the **Universal-free core** of Universal's dense iterative-refinement study into MTL5 as the dense sibling of `sparse::iterative_refine`. This is the first library piece of the Universal->MTL5 linear-algebra extraction (Universal epic stillwater-sc/universal#1204, phase stillwater-sc/universal#1207); the posit/precision *experiments* that drive it live in the new `mp-ir` composition layer, keeping MTL5 free of any Universal dependency.

## What's added
- **`mtl::lu_iterative_refine<Working>(A, b, x, opt)`** (`operation/lu_iterative_refine.hpp`) -- factor A once in a low `Working` precision via `lu_factor`/`lu_solve`, then correct with a residual formed in the (deduced, higher) `Residual` precision:
  ```
  factor A ~= P L U   (Working)         r  = b - A x   (Residual)
  x = U\(L\ P b)      (Working)         d  = U\(L\ P r) (Working);  x += d
  ```
  Generic over the arithmetic type, no Universal dependency. Mirrors the sparse sibling's `refine_options`/`refine_result`: **best-iterate return**, **patience** for a noisy low-precision residual, `rel_tol` convergence, and an opt-in **`scaled`** (scale-and-round) variant that carries the correction magnitude in `Residual` precision so narrow-exponent `Working` factors don't underflow.
- **`mtl::normwise_backward_error(A, x, b)`** (`operation/backward_error.hpp`) -- `||b - A x||_inf / (||A||_inf ||x||_inf + ||b||_inf)` (Higham, *Accuracy and Stability of Numerical Algorithms*, Thm 7.1), the natural refinement quality/termination metric.
- Both added to the `mtl.hpp` umbrella.

Ported (Universal-free) from Universal `blas/ext/solvers/luir.hpp` (`SolveIRLU`) + `blas/utes/nbe.hpp`.

## Design notes
- Two precisions (`Working` factor + `Residual`), matching how the existing sparse `iterative_refine` already works (its factorization carries its own precision). Universal's 3-precision `SolveIRLU<High,Working,Low>` collapses here to factor-precision vs residual-precision; the experiment layer chooses the type combos.
- All norms accumulate in `double` so the residual yardstick is precision-independent.

## Tests (`op_test_lu_iterative_refine`, 18 assertions)
- same-precision exact solve;
- **float factor + double residual recovers accuracy** on a non-float-exact system: beats the plain float LU solve and reaches the double residual floor;
- `rel_tol` / `converged`; best-iterate never worse than the plain solve;
- `normwise_backward_error` tiny for the exact solution, O(1) for a wrong one.

Green on gcc (ctest); headers clean under clang `-Werror -Wall -Wextra -Wpedantic`.

Part of the Universal->MTL5 BLAS extraction (stillwater-sc/universal#1207).

## Summary by CodeRabbit

- **New Features**
  - Added dense mixed-precision LU iterative refinement with configurable stopping criteria and an optional scaled-residual mode.
  - Added a normwise backward error utility to quantify solution quality and expose it via the main umbrella include.

- **Bug Fixes / Robustness**
  - Added input validation for backward error (checks matrix squareness and matching vector sizes; throws on mismatch).

- **Documentation**
  - Expanded the unreleased changelog with definitions and intended numerical behavior for both operations.

- **Tests**
  - Added unit tests covering refinement accuracy/convergence, best-iterate selection, and backward-error correctness and error handling.


Ravenwater:
Addressed all review findings in ab0077d (verified on GCC 13 and Clang 18, 7 cases / 27 assertions).

| Finding | Resolution |
|---------|-----------|
| `backward_error.hpp` missing `<algorithm>` | Added the explicit include (was transitive). |
| `backward_error.hpp` no square/size validation | Added `std::invalid_argument` checks for square A and matching x/b lengths, mirroring `lu_iterative_refine`. |
| `lu_iterative_refine` residual computed twice | Norm the already-built `r` (in double) instead of a second matvec — one matvec/step, single source of truth for the residual. |
| test units mismatch (normalized vs absolute) | `res.rel_residual` is `best_rn/‖b‖∞` (=3 for `build(10)`); now recover the absolute best residual before comparing to the absolute plain residual, restoring the exact "never worse than plain" bound. |
| missing edge cases | Added 1×1 and empty-system tests for `lu_iterative_refine`, plus 1×1 / non-square / size-mismatch tests for `normwise_backward_error`. |

All valid; none skipped.

Ravenwater:
Added the follow-up in the latest push — direct rectangular and size-mismatch `REQUIRE_THROWS_AS` coverage for `lu_iterative_refine` itself (previously only `normwise_backward_error` had it), guarding its square-input validation from regression. Verified on GCC 13 and Clang 18 (29 assertions).

Issue #254, opened 2026-07-22
Extend accumulator-aware mult()/dot() support across remaining ITL Krylov solvers


Gurleen-kansray:
# Tracking: Extend `accumulator_traits` support to the remaining ITL Krylov solvers

## Background

`cg.hpp` was recently updated to route its matrix-vector products and inner products through the accumulator-aware interfaces (`mtl::mult<Accumulator>()` and `mtl::dot<Accumulator, value_type>()`) instead of relying on `A * x` and the default accumulation path.

This allows Conjugate Gradient (CG) to take advantage of `accumulator_traits`, enabling mixed-precision and quire-based accumulation for supported number systems.

Following that change, the remaining ITL Krylov solvers were audited to determine whether the same gap exists elsewhere.

---

## Root Cause

`operator*(Matrix, Vector)` in `include/mtl/mat/operators.hpp` (lines 85–100) is **not** routed through `mtl::mult()`. Instead, it implements its own accumulation loop:

```cpp
template <Matrix M, Vector V>
    requires (!Matrix<V>)
auto operator*(const M& A, const V& x)
{
    ...
    using result_t =
        std::common_type_t<typename M::value_type,
                           typename V::value_type>;

    ...

    auto acc = math::zero<result_t>();
    for (...) {
        acc += A(r, c) * x(c);
    }

    ...
}
```

As a result:

- No `Accumulator` template parameter is available.
- No `accumulator_traits` specialization is used.
- No BLAS/SIMD/native-fast-GEMM dispatch is available, since that logic exists only in `mtl::mult()` / `detail::mult_generic()`.

Consequently, any solver using `A * x` bypasses the accumulator-aware implementation, regardless of the accumulation type requested by the caller.

---

## Proposed Fix

Update each affected solver to follow the same implementation pattern already applied in `cg.hpp`:

- Replace `A * x` with a pre-allocated output vector and call:

  ```cpp
  mtl::mult<Accumulator>(A, x, Ax);
  ```

- Add

  ```cpp
  typename Accumulator = void
  ```

  to the solver template parameter list.

- Route inner products through

  ```cpp
  mtl::dot<Accumulator, value_type>(...)
  ```

  wherever appropriate.

This preserves existing behavior for default arithmetic while enabling accumulator-aware execution for mixed-precision and quire-enabled types.

---

## Planned Work

- [x] Update `bicgstab.hpp`
- [x] Update `gmres.hpp`
- [x] Update `idr_s.hpp`

[...]

Issue #18, opened 2026-04-04
epic: Aggregate types — ndarray, tensor, and Python/NumPy/SciPy integration


Ravenwater:
## Epic: Aggregate Types

Implement the aggregate type hierarchy described in [docs/architecture/aggregate-types.md](https://github.com/stillwater-sc/mtl5/blob/main/docs/architecture/aggregate-types.md), adding `ndarray<T, N>` and `tensor<T, Rank, Dim>` alongside the existing `vec::` and `mat::` types.

## Motivation

MTL5 currently provides vectors (rank-1) and matrices (rank-2). This epic adds:
- **N-dimensional arrays** for general multi-dimensional computation
- **True tensors** with Einstein notation for physics and engineering applications
- **Interop bridge** so existing vector/matrix types work with ndarray algorithms

Python bindings and SciPy integration are out of scope for this repo — they belong in the separate `mtl5-python` peer repository. The C++ types are designed to be "Python-ready" (expose `data()`, strides, contiguous layout checks) so that pybind11 bindings in mtl5-python are straightforward.

## Implementation Plan

| Step | Issue | Status |
|---|---|---|
| 1. ndarray core | #19 | Done |
| 2. Vector/matrix interop | #20 | Done |
| 3. tensor core | #22 | Done |

~~4. Python bindings~~ → moved to `mtl5-python` repo (see #21)
~~5. SciPy integration~~ → moved to `mtl5-python` repo (see #23)

## Acceptance Criteria

- [x] `ndarray<T, N>` supports shape, strides, element access, views, broadcasting (#19)
- [x] Existing vector/matrix types satisfy `NdArray` concept and work with generic ndarray algorithms (#20)
- [x] `tensor<T, Rank, Dim>` supports Einstein notation via expression templates (#22)
- [ ] ~~Python module provides zero-copy NumPy ↔ MTL5 conversion~~ → mtl5-python
- [ ] ~~SciPy sparse matrices and solvers can interop with MTL5 types~~ → mtl5-python

Ravenwater:
All C++ work complete: ndarray (#19), vector/matrix interop (#20), tensor (#22). Python/SciPy items moved to mtl5-python repo.

PR #280, opened 2026-07-25
fix(mat): transposed_view dangling reference in operator()

### LINKED ISSUE #279: transposed_view: dangling reference in operator()
## Bug
transposed_view::operator() declares const_reference as its return type but binds it to ref_(c, r), where compressed2D::operator() returns value_type by value for sparse accessors. This produces a reference to a temporary that dangles immediately.

## Impact
SIGSEGV the first time a caller exercises trans(A) through mult(). No prior solver (cg, bicgstab, gmres, idr_s) multiplies by trans(A), so this went undetected until BiCG's adjoint matvec exercised it.

## Fix
operator() should return value_type by value, matching the underlying matrix's own accessor semantics.

#### Found while
Wiring BiCG's accumulator support, part of the #254 Krylov coverage tracking issue.

Gurleen-kansray:
## Summary
transposed_view::operator() declared const_reference as its return type but bound it to ref_(c, r), where compressed2D::operator() returns value_type by value for sparse accessors. This produced a reference to a temporary that dangled immediately.

## Bug
SIGSEGV the first time a caller exercised trans(A) through mult() -- BiCG's adjoint matvec (stillwater-sc/mp-iterative#18) is the first solver to multiply by trans(A); cg/bicgstab/gmres/idr_s never triggered this path (verified: no trans( calls in any of those four files).

## Fix
operator() now returns value_type by value, matching the underlying matrix's own accessor semantics.

## Verification
Full regression suite: 144/144 tests passing.
itl_test_bicg (blocked on this fix): now passes, was SIGSEGV before.

## Related
Closes #279. Found while wiring BiCG's accumulator support, part of #254.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved transposed matrix view element access to safely handle sparse matrices that return temporary values.
  * Prevented potential dangling references when reading elements through a transposed view.
  * Preserved existing access behavior while returning stable values for callers.

### DIFF
--- a/include/mtl/mat/view/transposed_view.hpp
+++ b/include/mtl/mat/view/transposed_view.hpp
@@ -19,7 +19,14 @@ public:
 
     explicit transposed_view(const Matrix& m) : ref_(m) {}
 
-    const_reference operator()(size_type r, size_type c) const {
+    // NOTE: returns value_type by value, not const_reference. The underlying
+    // Matrix::operator() (e.g. compressed2D) returns value_type by value for
+    // sparse accessors (structural zeros have no storage to reference), so
+    // binding a const_reference to ref_(c, r) here produced a dangling
+    // reference to a temporary -- SIGSEGV the moment BiCG's trans(A) path
+    // exercised it through mult(). CG/BiCGSTAB/GMRES never triggered this
+    // since none of them multiply by trans(A).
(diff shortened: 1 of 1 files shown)


Ravenwater (approved):
ship it

Issue #297, opened 2026-07-26
perf: extend on-node threading to factorizations, triangular solve, and multi-loop GEMM


Ravenwater:
## Context
Follow-up to the on-node threading epic #221, whose **acceptance criteria are met** and which is now closed: the persistent `detail::thread_pool` shipped and the highest-value kernels run on it — blocked GEMM (#239), GEMV/`axpy`/`scal` (#240), `dot`/`nrm2`/col-major GEMV (#241), and sparse CRS SpMV (#242), which the Krylov/multigrid solvers inherit — with a documented model + scaling writeup (#243), bit-identical across thread counts, and unchanged at `MTL5_NUM_THREADS=1`.

#221 listed further kernel coverage in its "Scope" beyond the acceptance bar. This issue tracks that remainder so it isn't lost in a closed epic.

## Remaining scope (all currently serial)
- [ ] **Dense factorizations** — LU / QR / Cholesky (`operation/lu.hpp`, `qr.hpp`, `cholesky.hpp`). Parallelize the trailing-submatrix / panel updates.
- [ ] **Sparse factorization panel updates** — supernodal LU/LDLᵀ (`sparse/factorization/supernodal_*.hpp`): parallelize the dense block updates within a supernode.
- [ ] **Sparse triangular solve** — level scheduling (`sparse/factorization/triangular_solve.hpp`) so independent rows within a level solve concurrently.
- [ ] **BLIS-style multi-loop GEMM** — today only the outer loop of the blocked GEMM is parallel; add parallelism across additional loops (jc/ic) for better 8-core+ efficiency.
- [ ] **General element-wise sweeps** — a `parallel_for` path for the element-wise matrix/vector expressions (partial today via `mat/operators.hpp`).

## Constraints (carry over from #221)
- Reuse the existing `detail::thread_pool` (`parallel_for` / `parallel_reduce`); no new per-iteration spawn.
- Determinism: keep results bit-identical where the serial path is (reductions may differ per thread count, as already documented).
- `MTL5_NUM_THREADS=1` (and unset) must be byte-identical to the serial path.

## Acceptance
- Each newly threaded kernel shows measurable multi-core scaling on `benchmarks/run_scaling.sh` and is exercised by an existing or added test.
- No behavior change at `MTL5_NUM_THREADS=1`.

## References
- Parent epic (closed, criteria met): #221
- Delivered batches: #239, #240, #241, #242, #243

Ravenwater:
[...]

PR #298, opened 2026-07-26
perf(operation): thread dense LU + Cholesky factorizations (#297 batch 1)

Ravenwater:
## Summary
First batch of extending on-node threading beyond the BLAS kernels (#297). Parallelizes the two dense factorization inner updates on the existing persistent `detail::thread_pool`:
- **LU** — the trailing-submatrix update (rows `i>k`) is partitioned over rows. Each row is written by exactly one chunk and reads only the shared pivot row `k`, so the contiguous chunking is **bit-identical to serial**.
- **Cholesky** — the column update (rows `i>j`) is partitioned over rows; each reads only already-finalized columns `k<j` and the shared column `j` → **bit-identical**.

Both use `parallel_for` with a 64K-work-per-chunk grain (the GEMV/`mult` convention), so the team is a **no-op at `MTL5_NUM_THREADS=1`** (default) — serial behavior and codegen unchanged. The LAPACK dispatch path (column-major float/double) is untouched; this threads the generic path (row-major / custom types).

## Changes
- `include/mtl/operation/lu.hpp` — parallelize the trailing update; `#include <detail/thread_pool.hpp>`
- `include/mtl/operation/cholesky.hpp` — parallelize the column update; includes
- `tests/unit/operation/test_factorization_mt.cpp` — new: sizes the pool to 4 before first use and asserts the library factorization is **bit-identical (`==`)** to an in-test serial reference (512×512 LU, 768×768 Cholesky — 557,956 assertions), plus residual checks

## Test Results
| Target | gcc build | gcc test | clang build | clang test |
|--------|-----------|----------|-------------|------------|
| op_test_factorization_mt | OK | PASS (557,956 assertions, threaded on 20 cores) | OK | PASS |
| op_test_lu | OK | PASS | OK | PASS |
| op_test_cholesky | OK | PASS | OK | PASS |
| op_test_lu_iterative_refine | OK | PASS | — | — |

Also **race-free under ThreadSanitizer** (`-DMTL5_SANITIZE=thread`, clang) — the MT test runs clean.

## Scope
Batch 1 of #297. Remaining items stay open there: QR/Householder, supernodal sparse factorization panels, sparse triangular-solve level scheduling, BLIS-style multi-loop GEMM, general element-wise sweeps.

## Test plan
- [ ] Promote to ready when satisfied: `gh pr ready`

Relates to #297

## Summary by CodeRabbit

- **Performance**
  - Improved LU and Cholesky factorization speed by parallelizing the lower-triangular/off-diagonal and trailing-submatrix updates on multicore systems.

- **Bug Fixes**
  - Maintained identical factorization results and solve accuracy when switching from single-threaded to threaded execution.

- **Tests**
  - Added multithreaded regression tests that compare LU and Cholesky against in-test serial “golden” references with bit-for-bit equality, including coverage for small sizes (single-chunk boundary).

### DIFF
--- a/include/mtl/operation/cholesky.hpp
+++ b/include/mtl/operation/cholesky.hpp
@@ -2,13 +2,16 @@
 // MTL5 -- Cholesky factorization for symmetric positive definite matrices
 // A = L*L^T. In-place: lower triangle of A is overwritten with L.
 // Optional LAPACK dispatch when MTL5_HAS_LAPACK is defined and types qualify.
+#include <algorithm>
 #include <cassert>
 #include <cmath>
+#include <cstddef>
 #include <mtl/concepts/matrix.hpp>
 #include <mtl/concepts/vector.hpp>
 #include <mtl/vec/dense_vector.hpp>
 #include <mtl/math/identity.hpp>
 #include <mtl/interface/dispatch_traits.hpp>
+#include <mtl/detail/thread_pool.hpp>
 #ifdef MTL5_HAS_LAPACK
 #include <mtl/interface/lapack.hpp>
 #endif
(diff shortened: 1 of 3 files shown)


Ravenwater:
Addressed the review nitpick in the latest push: added 1×1 and 2×2 boundary cases for both threaded factorizations, exercising the `parallel_for` serial-fallback branch where `trailing`/`rows` collapses. Used explicit deterministic tiny matrices (so SPD / diagonal-dominance never depends on the RNG) and asserted bit-identity to the serial reference. Verified on GCC 13 and Clang 18.

PR #300, opened 2026-07-26
perf(operation): thread Householder reflector application (#297 batch 2)

Ravenwater:
## Summary
Batch 2 of the on-node threading rollout (#297): parallelizes the two Householder reflector primitives — the O(m·n) hot loops of the Householder-based dense factorizations.

- **`apply_householder_left`** — partitioned over columns `j`. Each column reads the shared reflector `v` and only column `j` of `A`, writing only column `j`, so contiguous chunking is **bit-identical to serial**. Speeds `qr_factor`/`qr_extract_Q` and the other left-application callers (`lq`, `hessenberg`, `eigenvalue_symmetric`).
- **`apply_householder_right`** — partitioned over rows `i` (same independence argument) — speeds `lq`/`hessenberg`/`eigenvalue_symmetric`.

Both use `parallel_for` with a 64K-work-per-chunk grain; **no-op team at `MTL5_NUM_THREADS=1`** (serial behavior/codegen unchanged). The LAPACK QR path (column-major float/double) is untouched.

## Changes
- `include/mtl/operation/householder.hpp` — parallelize both reflector applications; add `<algorithm>`, `<cstddef>`, `<detail/thread_pool.hpp>`
- `tests/unit/operation/test_factorization_mt.cpp` — add Householder/QR MT cases

## Test Results
| Target | gcc | clang |
|--------|-----|-------|
| op_test_factorization_mt (1,082,272 assertions, threaded on 20 cores) | PASS | PASS |
| op_test_qr | PASS | PASS |
| op_test_svd | PASS | PASS |
| op_test_eigenvalue | PASS | — |
| op_test_eigen_symmetric | PASS | PASS |

The MT test asserts both reflector primitives are **bit-identical (`==`)** to an in-test serial reference (512×512), a 400×300 QR reconstructs `A = Q·R`, and 1×1/2×2 boundary cases hit the serial-fallback branch. **Race-free under ThreadSanitizer** (`-DMTL5_SANITIZE=thread`, clang). The Householder-dependent factorization tests (QR/SVD/eigen) all pass unchanged, confirming the shared-primitive blast radius is safe.

## Scope
Batch 2 of #297. Remaining: supernodal sparse factorization panels, sparse triangular-solve level scheduling, BLIS-style multi-loop GEMM, general element-wise sweeps.

## Test plan
- [ ] Promote to ready: `gh pr ready`

Relates to #297

## Summary by CodeRabbit

* **Performance**
  * Improved matrix factorization performance by parallelizing Householder reflection operations across rows and columns.
  * Better utilizes available CPU resources for larger matrix workloads.

* **Reliability**
  * Preserved existing numerical results while validating parallel Householder operations and QR factorization.
  * Added coverage for small matrices and single-chunk execution scenarios.

### DIFF
--- a/include/mtl/operation/householder.hpp
+++ b/include/mtl/operation/householder.hpp
@@ -68,8 +68,10 @@ void apply_householder_left(M& A, const vec::dense_vector<T>& v, T beta,
     // only column j of A, writes only column j), so partitioning the columns
     // across the thread pool is bit-identical to serial. No-op at
     // MTL5_NUM_THREADS=1.
+    // Empty or beyond-end target column: no work (matches the serial loop, and
+    // avoids an unsigned underflow in n - col before deriving the work range).
+    if (col >= n) return;
     const size_type ncols = n - col;
-    if (ncols == 0) return;
     const std::size_t grain = std::max<std::size_t>(
         std::size_t{1}, std::size_t{65536} / (vlen ? static_cast<std::size_t>(vlen) : std::size_t{1}));
     detail::thread_pool::instance().parallel_for(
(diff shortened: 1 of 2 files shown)


Ravenwater:
Fixed the Major finding in the latest push. Correct catch — my refactor derived the work range as `n-col`/`m-row` on unsigned `size_type`, so a beyond-end offset (`col>n`/`row>m`) would underflow and drive `parallel_for` out of bounds, whereas the original serial loops did nothing. Added early returns on `col >= n` / `row >= m` (also covers the exact target-empty case), plus a regression asserting the output is unchanged for target-empty and beyond-end offsets. Current callers never pass an out-of-range offset, but the primitive is public and was previously safe. Verified GCC 13 + Clang 18.

PR #301, opened 2026-07-27
perf(sparse): level-scheduled parallel lower triangular solve (#297 batch 3 pilot)

Ravenwater:
## Summary
Batch 3 of the on-node threading rollout (#297): parallelize the sparse triangular solve — on the hot path of every sparse direct solve, previously fully serial. **Pilot scoped to `sparse_cholesky`'s forward solve** (per maintainer decision), to validate the approach in review before rolling out to the other solvers.

## Why it needs a new form
The existing `dense_lower_solve` is a **CSC column-scatter** (`x[i] -= L(i,j)·x[j]`): two independent columns write the same `x[i]` and their subtraction order isn't fixed, so it can't be parallelized deterministically. This adds a **row-oriented gather** that is both parallel and **bit-identical to serial**:

- **`build_lower_solve_schedule(L)`** — transpose `L`'s strictly-lower part to CSR so each row lists `(k, L(i,k))`, `k<i`, in **increasing k order** (the same order the CSC scatter accumulates them), and partition rows into dependency levels (`level[i] = 1 + max level of its deps`). O(nnz), **built once and reused** across solves (factor-once / solve-many).
- **`level_scheduled_lower_solve(schedule, x)`** — for each level, solve its rows in parallel. Rows in a level are mutually independent — each writes its **own** `x[i]`, reading only earlier levels → **race-free**, and the identical accumulation order makes it **bit-identical to `dense_lower_solve`**. No-op team (serial, byte-identical) at `MTL5_NUM_THREADS=1`.

## Changes
- `include/mtl/sparse/factorization/level_schedule.hpp` — new: schedule + parallel solve
- `include/mtl/sparse/factorization/sparse_cholesky.hpp` — forward solve uses the level-scheduled path; schedule cached lazily in `cholesky_numeric`. Backward transpose solve stays serial (follow-up).
- `tests/unit/sparse/test_trisolve_level_schedule_mt.cpp` — new MT test

## Test Results
| Target | gcc | clang |
|--------|-----|-------|
| sparse_test_trisolve_level_schedule_mt (130,713 assertions) | PASS | PASS |
| sparse_test_sparse_cholesky | PASS | PASS |
| full sparse suite (30 tests) | PASS | — |

Bit-identity (`==`) to `dense_lower_solve` is asserted across the level spectrum — block-diagonal, a fully sequential dense block, a **wide arrow level that splits across the pool at n=80000**, diagonal, and 1×1. **Race-free under ThreadSanitizer** (`-DMTL5_SANITIZE=thread`, clang).

## Scope / rollout
Pilot. Follow-ups after review: the backward transpose solve (`dense_lower_transpose_solve` — schedulable directly on the CSC via column levels), and wiring into `sparse_lu` / `sparse_ldlt` / `supernodal_lu` / `supernodal_ldlt`.

## Test plan
- [ ] Promote to ready: `gh pr ready`

Relates to #297

## Summary by CodeRabbit

* **New Features**
  * Added deterministic level-scheduled sparse forward substitution for lower-triangular solves, enabling safe parallelism within dependency levels.
* **Performance**
  * Sparse Cholesky now precomputes and reuses a forward-solve plan for faster repeated `solve()` calls.
  * Forward-solve results remain bit-for-bit consistent with the existing dense reference.
* **Tests**
  * Added unit tests for threaded level-scheduled solves (including empty, 1×1, block-diagonal, arrow-shaped, and diagonal cases) with exact equality checks.

### DIFF
--- a/docs-site/sync-content.mjs
+++ b/docs-site/sync-content.mjs
@@ -62,6 +62,7 @@ const FILE_MAP = {
   'position-mixed-precision-acceleration.md':   'design/mixed-precision-acceleration.md',
   'design/blas-kernel-architecture.md':         'design/blas-kernel-architecture.md',
   'design/multicore-scaling-investigation.md':  'design/multicore-scaling-investigation.md',
+  'design/parallelization-patterns-and-pitfalls.md': 'design/parallelization-patterns-and-pitfalls.md',
   'design/mixed-precision-custom-types-SIMD.md': 'design/mixed-precision-custom-types-SIMD.md',
 
   // ── Modernization ──────────────────────────────────────────────────────
--- a/docs/design/parallelization-patterns-and-pitfalls.md
+++ b/docs/design/parallelization-patterns-and-pitfalls.md
@@ -1,0 +1,121 @@
+# Parallelizing algorithms in MTL5: patterns and pitfalls
+
+A field guide distilled from the on-node threading rollout ([#297](https://github.com/stillwater-sc/mtl5/issues/297), building on the thread-pool epic [#221](https://github.com/stillwater-sc/mtl5/issues/221)). It records the parallelization *patterns* that worked, and — more valuably — the *correctness, caching, encapsulation, and exception-safety pitfalls* that code review surfaced, so the next algorithm we parallelize can reuse the insight instead of rediscovering it.
+
+The concrete case studies are:
+
+- **Batch 1 ([#298](https://github.com/stillwater-sc/mtl5/pull/298))** — dense LU trailing-submatrix update, Cholesky column update.
+- **Batch 2 ([#300](https://github.com/stillwater-sc/mtl5/pull/300))** — Householder reflector application (`apply_householder_left`/`_right`), which powers QR / LQ / Hessenberg / the symmetric eigensolver.
+- **Batch 3 pilot ([#301](https://github.com/stillwater-sc/mtl5/pull/301))** — sparse triangular solve via level scheduling.
+
+---
+
+## 1. The non-negotiable properties
+
+Every parallel kernel in MTL5 must hold these, in priority order:
+
+1. **Serial path unchanged.** At `MTL5_NUM_THREADS=1` (the default) the code runs the original serial path, byte-identical in behavior and codegen. The persistent `detail::thread_pool` makes this automatic: `parallel_for` runs a single `body(0, n)` when the team is size 1 or the work is below the grain, so *always calling `parallel_for`* is a zero-overhead no-op serially.
+2. **Bit-identical across thread counts** wherever the serial code is deterministic. Not just "close" — exactly `==`. This is what lets us assert equality in tests and makes threaded results reproducible.
+3. **Race-free**, proven under ThreadSanitizer.
+4. **Deterministic schedule.** The partition of work is a pure function of the input, so results never depend on scheduling timing.
+
+Determinism (2) is a *choice we design for*, not a given. See §3.
+
+---
+
+## 2. The two parallelism patterns
+
+### 2a. Embarrassingly parallel over independent outputs
+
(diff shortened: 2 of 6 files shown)


Ravenwater:
@coderabbitai review

Ravenwater:
Addressed all three findings in the latest push.

- **Major — stale cache (the important one):** correct catch. The schedule copied `L`'s values, but `cholesky_numeric::L` is public and can be re-factorized in place, so the cache could go stale. Reworked the schedule to be **value-agnostic** — it stores only the *positions* of `L`'s entries and reads the numbers from `L` at solve time. So an in-place same-pattern re-factorization (the transient/mp-spice path) is now picked up automatically, and the cache is rebuilt only when `L`'s pattern (`nnz`) changes (guarded in `solve()`, plus an assert that the schedule matches `L`). Still bit-identical to `dense_lower_solve`.
- **Major — RHS length:** added `assert(x.size() >= s.n)` in `level_scheduled_lower_solve`.
- **Minor — empty matrix:** added a `0x0` boundary case (zero-level build + no-op solve).

Verified GCC 13 + Clang 18, race-free under ThreadSanitizer, all 30 sparse tests pass.

Ravenwater:
@coderabbitai review

Ravenwater:
Addressed the escalated finding in the latest push.

**"nnz is not a pattern key" (the sharpened Major):** fully agree — nnz can't distinguish two same-size/same-nnz factors with different structure, and the guard/assert vanish in release. Rather than invent a run-time pattern key, I **build the schedule together with `L` in `sparse_cholesky_numeric` (at factorization time, while `L` is authoritative) and store it bound to the result**. That removes the lazy-build-from-mutable-`L` window and the weak key entirely — the schedule intrinsically belongs to the factorization. Value-agnostic still holds, so an in-place **same-pattern** re-factorization (mp-spice transient) is reflected automatically; a pattern change yields a new factorization with a new schedule. `cholesky_numeric` is documented as an immutable result (as all factorization results are).

The earlier 11:57 re-posts (RHS length / stale-cache-from-copied-values / empty-matrix) were against the pre-fix commit and are already addressed: the schedule is now value-agnostic (no copied values), `level_scheduled_lower_solve` asserts `x.size() >= s.n` (its public entry point, `cholesky_numeric::solve`, also validates sizes with a throw), and the `0x0` empty case is covered.

Verified GCC 13 + Clang 18, race-free under ThreadSanitizer, all 30 sparse tests pass.

Ravenwater:
@coderabbitai review

Ravenwater:
Enforced the invariant in the latest push (rather than just documenting it), taking your first suggested resolution — "make the coupled state private/read-only, or require all updates through a setter that rebuilds the schedule." I did both:

- `cholesky_numeric::L` and the schedule are now **private** (`L_`, `fwd_sched_`), settable only **together** via `set_factor(L)`, which rebuilds the schedule from `L_`. They can't drift out of sync, so there's no reliance on a run-time (nnz) key.
- A read-only `factor()` accessor replaces the public `L` member (one test line updated).

Blast radius was one external reader. Verified GCC 13 + Clang 18, race-free under ThreadSanitizer, all 30 sparse tests pass.

Ravenwater:
@coderabbitai review

Ravenwater:
Fixed the Critical `set_factor` finding in the latest push — good catch on both parts. (1) It now **validates** the factor is square and matches `symbolic.n` before any mutation (a mismatch would OOB in release where the asserts are gone). (2) It's now **strongly exception-safe**: the schedule is built into a local first, then `L_` and `fwd_sched_` are committed with noexcept moves, so a throw during schedule construction leaves the object unchanged (previously `L_` was replaced before the build, risking a new factor + stale schedule). Verified GCC 13 + Clang 18, TSan clean, all 30 sparse tests pass.

Ravenwater:
@coderabbitai review

Issue #221, opened 2026-07-19
perf: on-node threading beyond GEMM (persistent thread pool)


Ravenwater:
## Problem

On-node parallelism is the biggest measured performance gap. The **only**
multithreaded kernel today is the native blocked GEMM
(`include/mtl/detail/gemm_blocked.hpp`), which partitions one loop across a raw
`std::thread` team created/joined per outer iteration, with `MTL5_NUM_THREADS`
defaulting to **1**. Everything else is serial: GEMV, all L1 ops, all dense
factorizations, **all sparse factorizations**, and all iterative/Krylov/multigrid
solvers. Parallel efficiency of the current GEMM falls from ~92% (4 cores) to
~73% (8 cores) — 5.8x on 8 P-cores vs OpenBLAS 7.15x / MKL 7.30x.

## Goal

A reusable on-node threading layer, used across kernels rather than one bespoke
team inside GEMM.

## Scope

- A **persistent thread pool** abstraction (no per-iteration spawn/join), with a
  single place to choose the backend (`std::thread` pool now; leave room for
  OpenMP/TBB/`std::execution` later). Respect `MTL5_NUM_THREADS` / hardware
  concurrency.
- Re-target the blocked GEMM onto the pool and add BLIS-style multi-loop
  parallelism (not just the `ic` loop).
- Parallelize the embarrassingly-parallel kernels: GEMV, L1 ops (axpy/scal/dot
  reductions), element-wise ops.
- Parallelize where the dependency structure allows: supernodal factorization
  panel updates, sparse triangular solve (level scheduling), Krylov SpMV.
- Keep results deterministic / bit-reproducible where the current code is (the
  GEMM team already is).

## Acceptance criteria

- A documented thread-pool API; GEMM uses it with no per-iteration spawn.
- Measurable multi-core scaling improvement on the `benchmarks/run_scaling.sh`
  sweep (target: close the 8-core efficiency gap materially).
- No change to results with `MTL5_NUM_THREADS=1`.

**Effort:** moderate (pool + GEMM) to large (broad kernel coverage). Highest
immediate value. Part of the scale-out roadmap.

Ravenwater:
## Closing — acceptance criteria met

The on-node threading layer shipped across 5 merged PRs:

- Persistent `detail::thread_pool` (no per-iteration spawn); blocked GEMM re-targeted onto it — #239
- Row-major GEMV, `axpy`, `scal` on the pool — #240
[...]

PR #303, opened 2026-07-28
ci: add threaded + ThreadSanitizer test lanes (#299)

### LINKED ISSUE #299: ci: add a threaded test lane (MTL5_NUM_THREADS>1) to exercise the pool-based kernels
## Problem
The main CI (`​.github/workflows/ci.yml`) runs `ctest --test-dir build --output-on-failure` with **`MTL5_NUM_THREADS` unset**, so `detail::thread_pool::instance()` sizes to **1** and every kernel runs its **serial** path. The on-node threading built in #221 (and being extended in #297) is therefore almost never exercised in CI:

- The env-sized singleton paths — threaded `mult`/`gemm`, `gemv`, `dot`/`nrm2`, `axpy`/`scal`, sparse SpMV, and now the dense LU/Cholesky factorization updates — are **not** run multithreaded by any CI job.
- The only threaded coverage today is via workarounds: `test_gemm_mt` drives the low-level `gemm_blocked` with an explicit `nthreads`, and `test_factorization_mt` (#298) `setenv`s `MTL5_NUM_THREADS` before first pool use. Everything else is serial in CI.

So a race or a chunking bug in a pool-based kernel could ship green.

## Proposal
Add a CI job that runs the test suite with the pool actually multithreaded:

- **Threaded lane** — Linux GCC + Clang (the runners have multiple cores), same build as the default job, but `ctest` run with `MTL5_NUM_THREADS=4` in the environment. Fast (reuses the existing build; just re-runs ctest under the env var), so it can gate PRs like the other Tier-1 jobs.
- **Optional: ThreadSanitizer variant** — a `-DMTL5_SANITIZE=thread` build (Clang) running at least the threading-tagged tests (`[mt]` / `[threading]`) under `MTL5_NUM_THREADS=4`, to catch data races the plain run can miss. Can start scoped to the `[mt]` subset (`ctest -R` / Catch2 tag filter) to keep it cheap, then broaden.

## Notes
- Determinism: the deterministic kernels are bit-identical across thread counts, so a threaded run should produce the same pass/fail as serial — any divergence is a real bug the lane is meant to surface.
- Keep the existing serial `ctest` run too: the `MTL5_NUM_THREADS=1` byte-identical guarantee must stay covered.
- Once this lane exists, the per-test `setenv` trick in `test_factorization_mt` can stay (it keeps the test self-contained) but is no longer the only thing exercising the threaded path.

## Acceptance
- A CI job runs the full unit-test suite with `MTL5_NUM_THREADS=4` on Linux GCC + Clang, gating PRs.
- (Optional/stretch) a TSan lane runs the `[mt]`/`[threading]` tests race-clean.
- The default serial `ctest` run is retained.

## References
- Threading epic (delivered): #221 — batches #239–#243
- Threading extension (in progress): #297 — batch 1 #298
- Existing threaded tests: `tests/unit/operation/test_gemm_mt.cpp`, `tests/unit/operation/test_factorization_mt.cpp`

Ravenwater:
## Summary
The default `build` job runs `ctest` with `MTL5_NUM_THREADS` unset, so `detail::thread_pool` sizes to 1 and every pool-based kernel runs its **serial** path — the on-node threading from #221 and #297 was almost never exercised in CI (only the per-test `setenv` workaround in `test_factorization_mt` / the explicit-`nthreads` `test_gemm_mt` touched it). A race or chunking bug could ship green.

Adds two lanes:
- **`threaded`** — Linux GCC + Clang, same Release build as the default job, but `ctest` runs with **`MTL5_NUM_THREADS=4`** so the threaded `mult`/`gemv`/`dot`/`nrm2`/`axpy`/`scal`, sparse SpMV, dense LU/Cholesky factorization, Householder application, and sparse triangular-solve paths actually run multithreaded. Deterministic kernels are bit-identical across thread counts, so a divergence from serial is a real bug.
- **`tsan`** — Clang `-DMTL5_SANITIZE=thread` build of the `*_mt` tests, run under `MTL5_NUM_THREADS=4` to catch data races. Scoped to the `_mt` tests to stay cheap (add new `*_mt` targets as coverage grows).

Both run on **every push and PR, including drafts** (like `build`) — so threaded coverage no longer depends on the workaround, and it isn't subject to the draft-skip that the Tier-2 `regression` job has. The default serial `ctest` run is retained (the `MTL5_NUM_THREADS=1` byte-identical guarantee).

## Changes
- `.github/workflows/ci.yml` — new `threaded` and `tsan` jobs (between `lapack` and `regression`).

## Validation (local, since GH Actions can't run locally)
- YAML parses; jobs = `ascii-check, build, lapack, threaded, tsan, regression`.
- **Full suite passes with `MTL5_NUM_THREADS=4`** (151 tests) — what the `threaded` lane runs.
- The `*_mt` tests build under `-DMTL5_SANITIZE=thread` and pass under `MTL5_NUM_THREADS=4` (`ctest -R _mt`, 3 tests) — what the `tsan` lane runs.
- The real validation is the new jobs appearing green in **this PR's own checks** (they run on drafts).

## Test plan
- [ ] The new `threaded (g++-13)`, `threaded (clang++-18)`, and `tsan` checks pass on this PR
- [ ] Promote to ready when satisfied: `gh pr ready`

Closes #299

## Summary by CodeRabbit

* **Tests**
  * Added parallel Linux test coverage using multiple threads.
  * Added ThreadSanitizer checks for detecting data races in multithreaded tests.
  * Expanded continuous integration coverage with dedicated threaded and thread-safety test runs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -214,6 +214,98 @@ jobs:
       - name: Test
         run: ctest --test-dir build --output-on-failure
 
+  # -- Threaded lane: run the suite with the pool actually multithreaded ----
+  # Same Release build as the default job, but ctest runs with
+  # MTL5_NUM_THREADS=4 so detail::thread_pool sizes to a real team and the
+  # pool-based kernels (mult/gemv/dot/nrm2/axpy/scal, sparse SpMV, the dense
+  # factorization updates, Householder application, sparse triangular solve)
+  # actually run multithreaded -- the default job runs them serial. The
+  # deterministic kernels are bit-identical across thread counts, so pass/fail
+  # must match serial; a divergence is a real race/chunking bug this lane
+  # surfaces. Runs on every push and PR (incl. drafts), like the build job. (#299)
+  threaded:
+    name: Linux x64 threaded (${{ matrix.cxx }})
+    runs-on: ubuntu-24.04
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - cc: gcc-13
+            cxx: g++-13
+          - cc: clang-18
+            cxx: clang++-18
+
+    # Same trusted-run gate as the build job: forked PRs skip sccache so they
+    # cannot poison the shared GHA cache.
+    env:
+      SCCACHE_LAUNCHER: ${{ (github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository) && 'sccache' || '' }}
+
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+
+      - name: Set up sccache
+        if: >-
+          github.event_name != 'pull_request' ||
+          github.event.pull_request.head.repo.full_name == github.repository
+        uses: mozilla-actions/sccache-action@9e7fa8a12102821edf02ca5dbea1acd0f89a2696 # v0.0.10
+
+      - name: Configure
+        run: >
+          cmake -B build
+          -DCMAKE_CXX_STANDARD=20
+          -DCMAKE_C_COMPILER=${{ matrix.cc }}
+          -DCMAKE_CXX_COMPILER=${{ matrix.cxx }}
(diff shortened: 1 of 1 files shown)


Ravenwater:
@coderabbitai review

Issue #82, opened 2026-06-01
Epic: native dense BLAS performance — within 10–20% of hand-optimized OpenBLAS/MKL


Ravenwater:
## Goal

Make MTL5's **native** dense BLAS fast enough to be useful without an external library: **within 10–20% of hand-optimized OpenBLAS/MKL** for GEMM at practical sizes (and as close as the memory wall allows for GEMV/L1). Today the native path is a naive triple loop at ~4–5 GFLOP/s — ~15× off OpenBLAS/MKL (~70+ GFLOP/s) in our own sweeps (`benchmarks/data/`).

## Where we are (baseline)

- `detail::mult_generic` (`include/mtl/operation/mult.hpp`) is an unblocked IJK triple loop; `dot`/`two_norm` are simple scalar loops. No SIMD, no cache blocking, no packing.
- `dense2D`/`dense_vector` are contiguous (`.data()`), row/col-major via tags — but heap storage is **not over-aligned** and the element accessor recomputes the offset per call.
- CMake sets **no `-march`/SIMD flags**; there are no intrinsics or `#pragma omp simd` anywhere.
- Clean dispatch seam already exists: `mtl::mult`/`dot`/`two_norm` do `if constexpr (BlasDenseMatrix...) → BLAS, else → generic`. A native fast path slots in between BLAS and the naive fallback.

## The two levers (and the literature)

Performance comes from (1) **register/SIMD blocking** — keep an `mr×nr` C microtile in vector registers and accumulate `kc` rank-1 FMA updates; and (2) **cache blocking + packing** — the GotoBLAS/BLIS 5-loop nest (`kc→L1, mc×kc→L2, kc×nc→L3`) with operands **packed** into contiguous, aligned, SIMD-friendly buffers (the packing is really about minimizing TLB pressure, not just cache lines). Naive→tuned is a ~15–50× gap; these two levers close almost all of it.

Key design choices we adopt:
- **BLIS factoring**: write the 5 loops + packing **once** (portable templates); isolate the arch-specific work in a single **micro-kernel** customization point. One micro-kernel → all of level-3 later.
- **SIMD abstraction** (the "decouple loop unrolling from the algorithm" idea from xtensor/xsimd, Eigen `PacketMath`, and `std::simd`): write the kernel against a `batch`/packet type with a `constexpr` lane count and one `fma` primitive, so the same source targets SSE/AVX/AVX2/AVX-512/NEON.
[...]

PR #332, opened 2026-08-02
fix(bench): gate #82 on the median ratio with a per-size floor

### LINKED ISSUE #327: perf(gemm): epic #82 acceptance gate is marginal — native-fast sits at ~82% of OpenBLAS and fails at varying sizes between runs
## Summary

The #82 gate (`native-fast >= 80% of OpenBLAS for GEMM at N >= 256`) **fails on the i7-12700K**, but not for the reason first reported in #325. It is not a defect at one size. native-fast sits at roughly **82% of OpenBLAS on average with several points of run-to-run spread per size**, so it straddles the 80% threshold and whichever sizes happen to land low that run trip the gate.

## Evidence

Two runs of the identical protocol (sweep `65:1025:80`, one P-core, all vendor threading pinned to 1), same binaries, same machine:

| N | run A (committed CSVs) | run B (quiet re-run) |
|---:|---:|---:|
| 305 | 80.6% | **78.7%** LOW |
| 385 | 80.2% | **79.0%** LOW |
| 465 | **76.3%** LOW | 82.5% |
| 545 | 80.2% | 86.0% |
| 625 | 86.5% | 85.9% |
| 705 | 82.6% | 82.8% |
| 785 | 84.6% | 83.5% |
| 865 | 83.9% | 82.5% |
| 945 | 83.9% | 82.0% |
| 1025 | 82.1% | **79.2%** LOW |

Run A: `GATE FAIL ... 1/10 sizes (worst 76.3%)`.
Run B: `GATE FAIL ... 3/10 sizes (worst 78.7%)`.

**No size fails in both runs.** Per-size spread reaches 6 points (N=465: 76.3% vs 82.5%). The mean sits near 82% — inside the epic's stated 10–20% target — but the 80% gate has no margin, so noise decides pass/fail.

A fine sweep `433:505:8` on the quiet machine shows native-fast flat across the region originally flagged (55.8, 55.7, 58.6, 59.1, **58.7 at N=465**, 58.8, 58.7, 59.4, 59.3, 59.5 GFLOP/s) — no structure at 465.

## Correction to #325

`docs/benchmarks/i7-12700k.md` (merged) states the dip at N=465 is "reproducible, not measurement noise", citing three isolated repeats at 70.6/75.4/72.5%. **That check was invalid.** Those repeats were taken while the machine was under load — native-fast measured 38–40 GFLOP/s there versus 56.9 for the same size on an idle machine — and their ratios were compared against sweep-protocol ratios rather than against isolated-protocol runs at neighbouring sizes. The missing control (same protocol, neighbouring sizes, quiet machine) gives 82.1% at N=385 and 86.6% at N=545 against 79.5% at N=465, i.e. ordinary variation.

The page needs correcting; that is tracked as part of this issue.

## Why the variance

- **CPU governor is `powersave`** (`intel_pstate`). Clocks are not pinned, so per-size turbo behaviour varies between runs. Already listed as a caveat on the result page.
- The measurement takes the **median of a small iteration count** per size; at ~60 GFLOP/s and these matrix sizes each timing is short.

## Suggested actions

1. **Fix the result page** to describe a marginal gate with run-to-run variance rather than a size-specific dip.
2. **Decide what the gate should assert.** At ~82% mean with a few points of spread, an 80% single-run-single-size threshold is a coin flip. Options: gate on the *median across sizes*, require N consecutive failures, or lower the threshold and state the real margin.
3. **Reduce measurement variance** before treating any individual size as signal — `performance` governor, more iterations, and reporting a spread alongside the median.
4. Only once the above is stable is it worth asking whether specific sizes are genuinely slow.

## Note

The underlying performance claim in #325 is unaffected: native-fast is ~82% of OpenBLAS single-threaded and 70.3% at 8 threads. What changes is the attribution — this is a marginal gate and a noisy measurement, not a kernel defect at N=465.

Ravenwater:
Resolves #327.

## Summary

`run_gate` failed if **any** single size fell below the threshold — it gated on the minimum, the noisiest statistic available. Two runs of the identical protocol, same binaries and pinning, on the same idle machine, failed at **disjoint sets of sizes**, so the verdict was decided by noise rather than by the kernel.

| statistic | run A | run B | swing |
|---|---:|---:|---:|
| **median of ratios** | 82.3% | 82.5% | **0.2 pt** |
| mean | 82.1% | 82.2% | 0.1 pt |
| min (the old rule) | 76.3% | 78.7% | 2.4 pt |
| worst single size | — | — | **6.2 pt** |

The gate now asserts the **median** ≥ `--threshold` (0.80), with a new `--floor` (0.70) that no individual size may drop below. The median is ~30× more stable and matches what the epic actually claims — "within 10–20% of OpenBLAS" is an aggregate statement, not a per-size one. The floor keeps the rule honest: a genuine cliff still fails, while ±6-point noise does not.

## Changes

- `benchmarks/analyze_gate.py` — median-based gate, `--floor`, summary line (median/mean/min/max), per-size marks demoted to `ok`/`low`/`FLOOR` where only `FLOOR` is fatal. Validation rejects `--floor > --threshold`. The rationale and the numbers above are recorded in the docstring so the choice is not re-litigated from scratch.
- `benchmarks/README.md` — new "What the gate asserts, and why" section with the stability table.
- `benchmarks/data/README.md` — reproduction command updated.
- `docs/benchmarks/i7-12700k.md` — reported the gate as **FAILING**; now reports PASS at 82.3% with the aggregation rationale.

## Verification

| Check | Result |
|---|---|
| Run A (committed CSVs) | **PASS** — median 82.3%, exit 0 |
| Run B (independent re-run) | **PASS** — median 82.5%, exit 0 |
| Synthetic 50% cliff at N=705 | **FAIL** on floor, exit 1 |
| `--threshold 0.90` | **FAIL** on median |
| `--floor 0.9 --threshold 0.8` | rejected by argparse |
| `--reference blis` | PASS, median 84.1% |
| Table mode (non-gate path) | unchanged |
| `python3 -m py_compile` | clean |

The cliff case is the one that justifies the floor: the median stayed at 81.4% and would have passed on its own, but the floor caught it.

No gcc/clang build in the test matrix because **no C++ changed** — this PR touches one Python script and three Markdown files. `npm run build` for the docs site still produces 43 pages.

## One correction to the issue

#327 proposed "more iterations" as a variance fix. The data says otherwise: within-run stddev is already only **0.3–2.2%** of the median at 10 iterations. The variance is *between* runs (turbo/thermal state, allocation, page layout), not within them, so more iterations would cost time and change nothing. Recorded in the docstring and the README so it isn't retried. The remaining real lever is the `powersave` governor.

## Test plan

- [ ] CI passes
- [ ] Promote to ready when satisfied: `gh pr ready <N>`

## Summary by CodeRabbit

* **Documentation**
  * Updated benchmark acceptance criteria to require both an aggregate median performance threshold and a per-size minimum floor.
  * Clarified gate exit conditions, stability metrics, rerun variance, and dedicated-hardware guidance.
  * Updated benchmark results to show the GEMM gate passing with an 82.3% median ratio and 70% per-size floor.

* **Bug Fixes**
  * Improved benchmark gate evaluation with median and mean reporting, per-size floor checks, clearer diagnostics, and validation for incompatible thresholds.

### DIFF
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -212,25 +212,52 @@ BENCH_CPU=4 BENCH_SUITES=blas benchmarks/run_sweeps.sh 65:1025:80
 benchmarks/analyze_gate.py benchmarks/data/blas_sweep_native-fast.csv \
     benchmarks/data/blas_sweep_openblas.csv --peak-gflops 78
 
-# Pass/fail gate: GEMM native-fast >= 80% of OpenBLAS for N >= 256
+# Pass/fail gate: median GEMM ratio >= 80% of OpenBLAS for N >= 256,
+# and no individual size below the 70% floor
 benchmarks/analyze_gate.py benchmarks/data/blas_sweep_native-fast.csv \
-    benchmarks/data/blas_sweep_openblas.csv --gate --op gemm --threshold 0.80 --min-size 256
+    benchmarks/data/blas_sweep_openblas.csv --gate --op gemm \
+    --threshold 0.80 --floor 0.70 --min-size 256
 ```
 
-The gate (`--gate`) exits non-zero if native-fast drops below the threshold, so
-it can run in an opt-in workflow. It is **not** wired into the per-push CI:
-shared CI runners have unstable clocks and no P-core pinning, which makes
-absolute perf gates flaky. Run it on dedicated hardware.
+### What the gate asserts, and why
+
+`--gate` exits non-zero if **either**:
+
+- the **median** of the per-size ratios falls below `--threshold` (default 0.80), or
+- **any single size** falls below `--floor` (default 0.70).
+
+The median is the primary assertion because the epic's target — "within 10–20%
+of OpenBLAS" — is an aggregate claim, and because it is the only statistic here
+that reproduces. Two runs of the identical protocol on the same idle machine
+gave:
+
+| statistic | run A | run B | swing |
+|---|---:|---:|---:|
+| median of ratios | 82.3% | 82.5% | **0.2 pt** |
+| mean | 82.1% | 82.2% | 0.1 pt |
+| min (the pre-#327 rule) | 76.3% | 78.7% | 2.4 pt |
+| worst single size | — | — | **6.2 pt** |
+
+The old rule failed if *any* size was below threshold — gating on the minimum,
+the noisiest statistic available — and the two runs failed at disjoint sets of
+sizes. The median is ~30× more stable.
+
+The floor keeps the rule honest: a genuine cliff at one size must still fail, so
+no size may drop below `--floor`, set well under the threshold so ordinary
+±6-point noise cannot trip it.
+
+> **Raising the iteration count does not help.** Within-run stddev is already
+> only 0.3–2.2% of the median at 10 iterations — the variance is *between* runs
+> (turbo/thermal state, allocation, page layout), not within them.
+
+It is **not** wired into the per-push CI: shared runners have unstable clocks
+and no P-core pinning, which makes absolute perf gates flaky. Run it on
+dedicated hardware.
 
 **Measured results live on the per-system result pages, not here** — see
 [Intel i7-12700K](../docs/benchmarks/i7-12700k.md), indexed from
 [Benchmark systems](../docs/benchmarks/systems.md). Keeping numbers in one place
 means a re-run updates them once; this file documents how to *produce* them.
-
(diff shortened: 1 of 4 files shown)


Ravenwater:
Reopening to trigger the full matrix: the workflow's pull_request trigger has no explicit types, so ready_for_review does not re-run CI and Tier 2 regression was skipped on the draft runs.

PR #334, opened 2026-08-02
ci: run CI when a draft PR is promoted to ready

### LINKED ISSUE #333: ci: promoting a draft PR to ready never re-runs CI, so Tier 2 regression stays skipped
## Summary

`.github/workflows/ci.yml:7-8` declares:

```yaml
  pull_request:
    branches: [ main ]
```

With no `types:`, GitHub applies the default `[opened, synchronize, reopened]`. **`ready_for_review` is not among them**, so `gh pr ready` / clicking "Ready for review" triggers no workflow run.

The Tier 2 `regression` job gates on draft status (`ci.yml:315-317`):

```yaml
    if: >-
      github.event_name == 'pull_request' &&
      github.event.pull_request.draft == false
```

So for any PR opened as a draft, the only runs that exist were evaluated while `draft == true`, and `regression` skipped in all of them. Promoting to ready does not re-evaluate. **Tier 2 never runs unless someone pushes another commit afterwards.**

## Impact

`gh pr checks` reports the PR as fully green — the skipped job is not a failure — so this is silent. A reviewer sees green and merges believing the documented two-tier policy held, when Tier 2 (dense, sparse and iterative solvers at 100–50K DOF) never executed.

This is not hypothetical. On #332:

```
regression   skipping   0   .../job/91513952318      <- draft-era run
regression   skipping   0   .../job/91513898914      <- draft-era run
```

after `gh pr ready`. Closing and reopening the PR (which does fire `reopened`) produced the real result:

```
regression   pass   5m45s   .../job/91514867127
```

Same commit, same content — the only difference was whether a qualifying event had fired.

The `/resolve-issue` skill **always** opens PRs as drafts, so every PR created through it is affected by default.

## Fix

```yaml
  pull_request:
    branches: [ main ]
    types: [ opened, synchronize, reopened, ready_for_review ]
```

Adding `ready_for_review` makes promotion trigger a run in which `draft == false`, so Tier 2 evaluates as documented. The other three must be listed explicitly once `types:` is present, since specifying it replaces the default rather than extending it.

Optionally also add `converted_to_draft`, if there is value in re-running when a PR moves the other way; probably not worth the minutes.

## Verification

After the change, open a draft PR, run `gh pr ready`, and confirm a new CI run appears with `regression` in a non-skipped state.

## Note

Found while merging #332. Every PR merged in the current session had Tier 2 genuinely passing, because they were opened non-draft or received a later push — the gap only bites the draft-then-promote path.

Ravenwater:
Resolves #333.

## Problem

`pull_request` declared no `types`, so GitHub applied the default `[opened, synchronize, reopened]`. **`ready_for_review` is not in that set**, so promoting a draft triggered no run at all.

The Tier 2 `regression` job gates on `pull_request.draft == false`. For a PR opened as a draft, every run that exists was evaluated while it *was* a draft, so `regression` skipped in all of them — and promotion re-evaluated nothing. Tier 2 therefore never ran unless someone pushed another commit.

It failed silently: `gh pr checks` reports green, because a skipped job is not a failure. A reviewer sees green and merges believing the documented two-tier policy held.

Observed directly on #332 — same commit, same content:

```
regression   skipping   0       <- after `gh pr ready`
regression   pass       5m45s   <- after close/reopen (which fires `reopened`)
```

## Fix

```yaml
  pull_request:
    branches: [ main ]
    types: [ opened, synchronize, reopened, ready_for_review ]
```

The three defaults are listed explicitly because naming `types` **replaces** the default set rather than extending it — omitting them would silently stop CI on ordinary pushes to a PR, which would be far worse than the bug being fixed.

## Validating the fix on itself

`pull_request` workflows run from the PR's head branch, so this change is in effect for this PR. That makes it self-testing, which is why **this PR is deliberately opened as a draft**:

1. Draft runs → `regression` skips (the pre-fix behaviour, still correct: Tier 2 is meant to skip on drafts).
2. `gh pr ready` → with this change, a fresh run fires with `draft == false`.
3. `regression` executes rather than skipping — without any push in between.

Step 3 is exactly what does not happen on `main` today. I will post the before/after check output on this PR once promoted.

## Test plan

- [ ] Draft CI passes, `regression` skips (expected on a draft)
- [ ] `gh pr ready` triggers a new run without a push
- [ ] `regression` runs and passes in that new run
- [ ] Merge

## Summary by CodeRabbit

* **Chores**
  * Continuous integration now runs when pull requests are opened, updated, reopened, or marked ready for review.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,6 +6,16 @@ on:
     tags: [ 'v[0-9]*' ]
   pull_request:
     branches: [ main ]
+    # `ready_for_review` matters: the Tier 2 regression job gates on
+    # `pull_request.draft == false`, so a PR opened as a draft only ever has
+    # runs in which that job skipped. Without this trigger, promoting the draft
+    # re-evaluates nothing and Tier 2 never runs unless another commit is
+    # pushed -- and `gh pr checks` still reports green, because a skipped job
+    # is not a failure (#333).
+    #
+    # The other three are the defaults, and must be listed explicitly: naming
+    # `types` REPLACES the default set rather than extending it.
+    types: [ opened, synchronize, reopened, ready_for_review ]
 
 jobs:
   # -- ASCII guard: source files must contain only ASCII bytes ---------


Ravenwater:
## Before/after evidence — the fix validated on itself

Because `pull_request` workflows run from the PR's head branch, this change is live for this PR. It was therefore opened as a **draft** on purpose, so promotion exercises exactly the path that is broken on `main`.

**No commit was pushed between the two observations.** The only thing that changed is the PR's draft state.

### Before — draft

```
40 pass
 2 skipping

regression   skipping   0   .../runs/30758883400/job/91526522429
regression   skipping   0   .../runs/30758894043/job/91526435177
```

Two runs existed (`push` + `pull_request`/opened), and `regression` skipped in both. That is correct behaviour for a draft — Tier 2 is *meant* to skip there.

### After — `gh pr ready`, no push

```
runs before promotion: 2
runs after promotion : 3

  30759199413  pull_request  in_progress   <-- fired by ready_for_review
  30758894043  pull_request  completed
  30758883400  push          completed
```

**A third run appeared with no commit pushed.** On `main` today this produces nothing at all: `ready_for_review` is not in the default `types` set, so promotion is not an event the workflow listens for, and the draft-era runs — in which `regression` already skipped — remain the only ones. `gh pr checks` then reports green, because a skipped job is not a failure.

That new run is the whole fix. Run `30759199413` is evaluating with `pull_request.draft == false`, so the Tier 2 `regression` job is no longer gated out.

### Still pending

`regression` has `needs: build`, so its job is not created until the build stage of run `30759199413` finishes. I will post the final `regression` result here once it lands — that is the last confirming detail, not the crux, which is the third run existing at all.

For contrast, this is what the same situation looked like on #332 *without* the fix, same commit and content:

```
regression   skipping   0       <- after `gh pr ready`
regression   pass       5m45s   <- only after a close/reopen, which fires `reopened`
```

Ravenwater:
## Final result — Tier 2 ran and passed on promotion

The pending piece has landed. Run `30759199413`, the one fired by `ready_for_review`:

```
regression   pass       5m51s   .../runs/30759199413/job/91527295831
regression   skipping   0       .../runs/30758883400/job/91526522429   <- the push run, still correctly skipped
```

Full check state after promotion: **41 pass, 1 skipping** (the remaining skip is `regression` on the `push`-event run, which is correct — that job requires `github.event_name == 'pull_request'`).

So the complete chain, with no commit pushed at any point:

| | runs | `regression` |
|---|---|---|
| draft | 2 | `skipping`, `skipping` |
| after `gh pr ready` | 3 | **`pass` 5m51s** |

Tier 2 now executes on draft promotion, which is exactly what does not happen on `main`. That is the behaviour #333 describes and this PR restores.

PR #339, opened 2026-08-02
fix(operation): replace the SVD iteration and harden Householder reflectors

### LINKED ISSUE #337: bug(operation): svd returns NaN or badly wrong singular values on ordinary symmetric input
`mtl::svd` / `mtl::detail::singular_values` returns all-`NaN` singular values for a large fraction of ordinary symmetric matrices, and materially wrong values for many of the rest. Everything layered on it — `condition_number`, `rcond`, `numerical_rank`, `nullity` — inherits the failure.

Found at `6eb9376` while binding the dense surface for Python.

## Reproducer

For a symmetric matrix, `sigma_max` equals the spectral radius by definition, which gives a self-contained check needing no reference implementation.

```cpp
#include <cstdio>
#include <cmath>
#include <random>
#include <algorithm>
#include <mtl/mat/dense2D.hpp>
#include <mtl/operation/operators.hpp>
#include <mtl/mat/operators.hpp>
#include <mtl/operation/svd.hpp>
#include <mtl/operation/spectral_properties.hpp>

int main() {
    std::mt19937 gen(0);
    std::normal_distribution<double> nd(0.0, 1.0);
    int nan_cases = 0, total = 0;
    double worst = 0.0; std::size_t worst_n = 0;

    for (int trial = 0; trial < 200; ++trial) {
        std::size_t n = 3 + (trial % 18);
        mtl::mat::dense2D<double> A(n, n);
        for (std::size_t i = 0; i < n; ++i)
            for (std::size_t j = i; j < n; ++j) { double v = nd(gen); A(i,j)=v; A(j,i)=v; }

        auto sv = mtl::detail::singular_values(A);
        ++total;
        bool has_nan = false;
        for (double s : sv) if (!std::isfinite(s)) has_nan = true;
        if (has_nan) { ++nan_cases; continue; }

        double smax = *std::max_element(sv.begin(), sv.end());
        double sr = mtl::spectral_radius(A);
        double rel = std::abs(smax - sr) / (sr > 0 ? sr : 1.0);
        if (rel > worst) { worst = rel; worst_n = n; }
    }
    std::printf("symmetric inputs: %d/%d produced NaN singular values\n", nan_cases, total);
    std::printf("worst |sigma_max - spectral_radius|/spectral_radius = %.3e (n=%zu)\n", worst, worst_n);
}
```

```
first NaN at n=10, singular values: nan nan nan nan nan nan nan nan nan nan
symmetric inputs: 61/200 produced NaN singular values
worst |sigma_max - spectral_radius|/spectral_radius = 1.427e+00 (n=9)
```

So roughly **30% of symmetric inputs return all-NaN**, and among those that return finite values the largest singular value can be off by 143%.

## Cross-checked against a reference

Comparing `singular_values` with LAPACK (via NumPy) on the same matrices, largest relative deviation:

| case | max rel Δ | mtl5 σ_min | reference σ_min |
|---|---|---|---|
| general n=4 | 1.2e-15 | 0.318272 | 0.318272 |
| general n=6 | 2.0e-02 | 0.388098 | 0.302946 |
| general n=8 | 2.9e-02 | 0.093346 | 0.221994 |
| symmetric n=4 | **NaN** | NaN | 0.048991 |
| symmetric n=8 | 4.5e-01 | 0.360228 | 0.360228 |
| symmetric n=20 | **NaN** | NaN | 0.272870 |
| SPD n=10 | 2.1e-02 | 10.164756 | 10.164756 |
| rect 10x6 | 1.9e-01 | 1.054970 | 1.054970 |

It is not confined to symmetric input, and it is not a tolerance question — some sizes are exact to 1e-15 while neighbouring sizes are wrong by tens of percent, which reads like a convergence or deflation failure rather than accumulated rounding.

Note the eigensolvers on the same matrices are fine: over 120 random matrices, `eigenvalue_symmetric` had worst relative error 3.7e-15 and `eigenvalue` 1.1e-9. So this looks specific to the SVD path rather than to the shared QR machinery.

## Blast radius

`condition_number`, `rcond`, `numerical_rank` and `nullity` all call `detail::singular_values`, so all four are affected. Concretely, `condition_number` reports 18.17 where LAPACK gives 25.83 on an 8×8 symmetric matrix — and returns NaN whenever the underlying SVD does.

## Suggested regression

The `sigma_max == spectral_radius` identity above needs no reference implementation and catches both failure modes. A NaN guard on the returned values would catch the more severe one on its own.

Reported from stillwater-sc/mtl5-python, where `svd`, `svdvals`, `condition_number`, `rcond`, `numerical_rank` and `nullity` are being withheld until this is fixed. The eigen bindings are shipping, since they check out.

Ravenwater:
Resolves #337.

### 1. Householder underflow — the NaN

`sigma = sum(x(i)^2)` was formed unscaled, and only **exact** zero was guarded. A merely tiny sigma reached

```cpp
beta = 2*v0*v0 / (sigma + v0*v0);   // both terms flushed to zero -> 0/0 = NaN
v(i) /= v0;                          // -> infinity
```

`qr_factor` *stores* `v` below the diagonal, so the poison spread into the factor and everything downstream. Fixed by scaling the column before squaring, returning a clean unit `v` when `beta` underflows to zero, and returning early from `apply_householder_left/right` when `beta == 0` — the reflection is the identity there, and `beta * v(i) * w` would otherwise evaluate `0 * inf = NaN`.

This is a shared primitive, so the hardening benefits QR and LQ too.

### 2. The SVD iteration — the wrong values

The alternating-QR sweep converged only linearly and then **corrupted its own answer**. On a 10×10 SPD case:

| iteration | worst singular value error | off-diagonal / diagonal |
|---|---|---|
| 10 | 5.8e-04 | 1.2e-02 |
| 100 | **8.4e-09** | 4.6e-06 |
| 1000 | **1.359** | 2.6e-27 |

The convergence test measured off-diagonal mass over diagonal mass, which **kept shrinking through the corruption** and so read as perfectly converged. `max_iter` was `100*max(m,n)` = 1000 here, so the corrupted value is exactly what was returned.

I checked whether a cheaper fix would do: a "diagonal has stabilized" criterion does **not** work either — the diagonal's relative change decreases monotonically *across* the failure (1.1e-10 at iter 120, 1.3e-11 at 140 where the error jumps to 0.1155). The failure is a discrete event, not drift, so there is no threshold to tune.

Replaced with **one-sided Jacobi** — which is what the file header already advertised. It orthogonalizes the columns by plane rotations, accumulates them into V, and reads the singular values off as the column norms. Quadratic convergence, no shifts or deflation, and no degenerate case to guard: an already-orthogonal pair is simply skipped. U is completed to a genuine orthogonal basis when the input is rank deficient or `m > n`.

## Verification against LAPACK

`dgesvd` over 1800 random matrices (general, symmetric, SPD; n = 2..16):

| | before | after |
|---|---|---|
| NaN cases | 63–124 per 600 | **0 per 600** |
| worst rel deviation | 1.8 – 5.8 (180–577%) | **3.0e-15** |

The issue's own reproducer goes from 61/200 NaN and 143% worst error to **0/200 NaN and 6.0e-15**.

Decomposition invariants over 600 cases spanning square, tall, wide, 1×N, N×1 and rank-deficient input:

| invariant | worst |
|---|---|
| `\|\|A - U*S*V^T\|\|` / `\|\|A\|\|` | 4.9e-15 |
| `\|U^T U - I\|` | 5.6e-16 |
| `\|V^T V - I\|` | 6.7e-15 |
| negative singular values | none |
| ordering / off-diagonal defect | none |

## API note

The default `tol` now means machine epsilon rather than `1e-10`, which is the right threshold for the Jacobi orthogonality test — the old default would have capped accuracy at ~1e-10. An explicit argument still overrides it.

## Tests

The existing cases use small or specially structured matrices and passed throughout, which is why this survived. Added:

- the issue's own **`sigma_max == spectral_radius`** identity for symmetric input, which needs no reference implementation;
- a reconstruction + orthogonality sweep over seven shapes including wide, tall and square;
- a rank-deficient case that specifically exercises the U completion path.

Two of the three fail without the fix (verified by reverting the headers and keeping the tests).

## Test results

| | GCC 13 | Clang 18 |
|---|---|---|
| `op_test_svd` | PASS (1841 assertions) | PASS |
| full suite | PASS 154/154 | PASS 154/154 |

## Note

Reported from `mtl5-python`, where `svd`, `svdvals`, `condition_number`, `rcond`, `numerical_rank` and `nullity` are all withheld pending this. Their `test_svd_is_not_exposed` asserts the absence of those symbols, so it will fail once they are bound — which is their intended reminder.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved numerical stability for Householder reflections, including tiny, zero, and identity vectors.
  * Enhanced SVD accuracy and robustness across empty, rectangular, and rank-deficient matrices.
  * Singular values are now reliably nonnegative and ordered, with improved orthogonal vector completion.
  * SVD reconstruction and orthogonality behavior is more consistent across matrix shapes.

* **Tests**
  * Added regression coverage for spectral properties, reconstruction accuracy, orthogonality, ordering, and rank-deficient inputs.

### DIFF
--- a/include/mtl/operation/householder.hpp
+++ b/include/mtl/operation/householder.hpp
@@ -23,34 +23,64 @@ std::pair<vec::dense_vector<T>, T> householder(const vec::dense_vector<T>& x) {
     const size_type n = x.size();
 
     vec::dense_vector<T> v(n);
+
+    // Scale by the largest magnitude before squaring. Forming sum(x(i)^2)
+    // directly underflows to zero once the entries are small -- and the old
+    // `sigma == 0` test only caught EXACT zero, so a merely tiny sigma fell
+    // through to `beta = 2*v0*v0 / (sigma + v0*v0)` with both terms flushed to
+    // zero, i.e. 0/0 = NaN, plus `v(i) /= v0` overflowing to infinity.
+    //
+    // The SVD's alternating-QR iteration drives exactly that: as it converges
+    // the trailing columns shrink toward zero, so the reflectors underflow and
+    // poison an already-correct answer with NaN (#337). Scaling keeps the
+    // squares O(1), so the only way sigma vanishes now is that x(1:) really is
+    // negligible against x(0) -- which is the genuine "already along e_1" case.
+    T scale = math::zero<T>();
+    for (size_type i = 0; i < n; ++i) {
+        const T axi = abs(x(i));
+        if (axi > scale) scale = axi;
+    }
+
     for (size_type i = 0; i < n; ++i)
         v(i) = x(i);
-
-    // Compute sigma = sum(x(1:end)^2)
-    T sigma = math::zero<T>();
-    for (size_type i = 1; i < n; ++i)
-        sigma += x(i) * x(i);
-
     v(0) = math::one<T>();
 
-    if (sigma == math::zero<T>()) {
-        // x is already along e_1
+    if (scale == math::zero<T>())
+        return {v, math::zero<T>()};   // x is the zero vector
+
+    // sigma = sum((x(i)/scale)^2) for i >= 1, computed in the scaled variables
+    T sigma = math::zero<T>();
+    for (size_type i = 1; i < n; ++i) {
+        const T xs = x(i) / scale;
+        sigma += xs * xs;
+    }
+
+    if (sigma == math::zero<T>())
+        return {v, math::zero<T>()};   // x is already along e_1
+
+    const T x0 = x(0) / scale;
+    const T norm_x = sqrt(x0 * x0 + sigma);
+    T v0;
+    if (x0 <= math::zero<T>())
+        v0 = x0 - norm_x;
+    else
+        v0 = -sigma / (x0 + norm_x);
+
+    const T beta = T(2) * v0 * v0 / (sigma + v0 * v0);
+
+    // beta == 0 means the reflection is the identity, so v carries no
+    // information -- and computing it would divide by a v0 small enough to
+    // overflow. Return a clean unit v: qr_factor STORES v below the diagonal,
+    // so letting huge or infinite entries through would poison the factor and
+    // every later use of it.
+    if (beta == math::zero<T>())
         return {v, math::zero<T>()};
-    }
 
-    T norm_x = sqrt(x(0) * x(0) + sigma);
-    if (x(0) <= math::zero<T>()) {
-        v(0) = x(0) - norm_x;
-    } else {
-        v(0) = -sigma / (x(0) + norm_x);
-    }
-
-    T beta = T(2) * v(0) * v(0) / (sigma + v(0) * v(0));
-
-    // Normalize v so that v(0) = 1
-    T v0 = v(0);
-    for (size_type i = 0; i < n; ++i)
-        v(i) /= v0;
+    // Normalize so that v(0) = 1. Both numerator and denominator are in the
+    // scaled variables, so the ratio is the same as before the scaling.
+    for (size_type i = 1; i < n; ++i)
+        v(i) = (x(i) / scale) / v0;
+    v(0) = math::one<T>();
 
     return {v, beta};
 }
@@ -71,6 +101,10 @@ void apply_householder_left(M& A, const vec::dense_vector<T>& v, T beta,
     // Empty or beyond-end target column: no work (matches the serial loop, and
     // avoids an unsigned underflow in n - col before deriving the work range).
     if (col >= n) return;
+    // beta == 0 is the identity reflection. Returning early is not just an
+    // optimisation: `beta * v(i) * w` would evaluate 0 * inf = NaN if any
+    // reflector entry had overflowed.
+    if (beta == math::zero<T>()) return;
     const size_type ncols = n - col;
     const std::size_t grain = std::max<std::size_t>(
         std::size_t{1}, std::size_t{65536} / (vlen ? static_cast<std::size_t>(vlen) : std::size_t{1}));
@@ -106,6 +140,8 @@ void apply_householder_right(M& A, const vec::dense_vector<T>& v, T beta,
     // Empty or beyond-end target row: no work (matches the serial loop, and
     // avoids an unsigned underflow in m - row before deriving the work range).
     if (row >= m) return;
+    // Identity reflection -- see the note in apply_householder_left.
+    if (beta == math::zero<T>()) return;
     const size_type nrows = m - row;
     const std::size_t grain = std::max<std::size_t>(
         std::size_t{1}, std::size_t{65536} / (vlen ? static_cast<std::size_t>(vlen) : std::size_t{1}));
--- a/include/mtl/operation/householder.hpp
+++ b/include/mtl/operation/householder.hpp
@@ -23,6 +23,7 @@ std::pair<vec::dense_vector<T>, T> householder(const vec::dense_vector<T>& x) {
     const size_type n = x.size();
 
     vec::dense_vector<T> v(n);
+    if (n == 0) return {v, math::zero<T>()};   // no element at index 0 to write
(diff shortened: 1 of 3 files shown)


Ravenwater:
Addressed all four findings in `0cee70f`. **Two were real defects in the new Jacobi code**, and I reproduced both before fixing rather than taking them on faith.

### 1. `householder`: guard `n == 0` — valid

`v(0)` was written unconditionally, out of bounds for an empty input. Not reachable from `qr_factor` (its column length is always ≥ 1), but `householder` is a public template. Guarded, with a test.

### 2. Scale the columns before forming the inner products — valid, and worse than described

`alpha`/`beta`/`gamma` are sums of squares, so taken raw they overflow near 1e200 and underflow near 1e-200. The underflow was the dangerous half: `scale == 0` made the code `continue`, **skipping the rotation entirely**, so the columns were never orthogonalized. Measured before the fix:

| scale | σ₀ | relative reconstruction | \|UᵀU − I\| |
|---|---|---|---|
| 1e-200 | **0** | **1.0** (total failure) | 0 |
| 1e-160 | 4.12e-160 | 0 | **0.87** |
| 1e+200 | **inf** | — | — |

After: σ₀ = 4.7453e±200 at every scale, reconstruction ≤ 7.3e-16, orthogonality ≤ 2.5e-16.

One scale shared by *both* columns leaves `zeta` — and therefore the rotation — unchanged; per-column scales would not. The final column norms are scaled the same way.

### 3. The fixed `0.5` completion bound — valid, and the analysis is exactly right

With `d` directions missing, the residuals of the `m` canonical vectors satisfy `Σ ||P_perp e_k||² = d`, so the best is only guaranteed to reach `sqrt(d/m)` — 0.25 for a single missing direction at m = 16. A `> 0.5` test rejects **every** candidate while the basis is still incomplete, leaving stale identity columns in `U`.

Constructed the case the comment implies — `A = I - v*vᵀ` with `v = ones/sqrt(m)`, so every canonical vector has residual exactly `1/sqrt(m)`:

| m | residual per candidate | \|UᵀU − I\| before | after |
|---|---|---|---|
| 4 | 0.500 | **0.87** | 2.2e-16 |
| 9 | 0.333 | **0.94** | 3.3e-16 |
| 16 | 0.250 | **0.97** | 4.4e-16 |

Note it failed at m = 4 too, where the residual is exactly 0.5 and `> 0.5` is false.

Replaced with pivoted Gram-Schmidt: take the candidate with the largest remaining residual each step. That maximum is at least `sqrt(d/m) > 0` whenever `d > 0`, so it always makes progress, and keeping the candidates orthogonalized incrementally holds the completion at O(m³).

### 4. Tolerance semantics — intentional, now documented in the header

Confirmed intentional. The old `1e-10` default would cap accuracy several orders short of what one-sided Jacobi achieves — it would have defeated the fix. The header now states this, so it is discoverable outside this PR, and callers can still pass `1e-10` explicitly for the old looser behaviour.

### Why the existing tests missed 2 and 3

Worth recording: the sweeps use unit-scale matrices, and my rank-deficient case (`A(i,j) = (i+1)(j+1)`) has a complement that aligns with the coordinate axes, so its residuals were far above any threshold. Added an extreme-scaling case spanning 1e-200..1e200, the `I - v*vᵀ` completion case at m = 4, 9, 16, and an empty-input Householder case.

### Re-verified after the changes

- LAPACK `dgesvd`, 1800 random matrices: **0 NaN**, worst deviation **2.7e-15**
- Invariants over 600 shapes: `||A - U*S*Vᵀ||` ≤ 4.2e-15, `|UᵀU − I|` ≤ 6.7e-16, `|VᵀV − I|` ≤ 2.4e-14
- Full suite: **154/154** on GCC 13 and Clang 18

Issue #330, opened 2026-08-02
ci: the Highway SIMD backend (simd/batch.hpp) is compiled by no CI job


Ravenwater:
## Summary

No workflow sets `MTL5_WITH_HIGHWAY=ON`, so `MTL5_HAS_HIGHWAY` is never defined in CI and the **Highway-backed `simd::batch` implementation is compiled by no runner**. Every CI job builds the portable scalar fallback instead.

This is the same class of gap as #329 (benchmark sources compiled nowhere), but on library code.

## Scope — narrower than it first appears

Worth stating precisely, because most of the fast-GEMM stack *is* covered and I initially assumed otherwise:

| Code | Covered in CI? | Why |
|---|---|---|
| `detail/gemm_blocked.hpp` | **yes** | included unconditionally by `mtl.hpp:18` and `operation/mult.hpp:20` |
| `#ifdef MTL5_NATIVE_FAST_GEMM` dispatch in `mult.hpp` (lines 217, 278, 345) | **yes** | `tests/unit/operation/test_gemm_blocked.cpp:12` and `test_fast_gemm_numeric.cpp:15` `#define MTL5_NATIVE_FAST_GEMM 1` before including it |
| `simd/batch.hpp` Highway path (lines 41–94) | **no** | requires `MTL5_HAS_HIGHWAY`, which only the CMake option defines |

So the blocked GEMM, the micro-kernel, and the `mult()` routing are all exercised. What is not is the ~54-line Highway specialization of `simd::batch` — the actual SIMD register type underneath those kernels. In CI it is always `size == 1` scalar.

Unlike the `NATIVE_FAST_GEMM` macro, a test cannot opt in by `#define`-ing `MTL5_HAS_HIGHWAY`: that path needs the Highway headers and library, which are absent unless the build fetches or finds them.

## Why it matters

The kernels are tested for numerical correctness against the scalar fallback only. A change to the Highway specialization — lane handling, masked edges, an FMA ordering difference — compiles and passes CI while being broken in every build users actually benchmark or deploy with, since `-DMTL5_WITH_HIGHWAY=ON` is what the `native-fast` variant and all published performance numbers use.

## Cost

[...]

Issue #321, opened 2026-08-02
bench(sparse): run_scaling_297.sh SPARSE_SIZES default (200,320) is intractable — 3-D Laplacian factorization dominates


Ravenwater:
## Summary

`run_scaling_297.sh` defaults to `SPARSE_SIZES=200,320` (`run_scaling_297.sh:34`). On an i7-12700K that default does not finish: a `T=1` run was killed after **3 h 28 min** without completing the first size, while the four dense families in the same pass finished in under 10 minutes combined.

The binary's own documented default is `100,160` (`bench_sparse.cpp:26`). The driver overrides it with values ~4x larger in DOF.

## Measured cost (whole `bench_sparse` suite, `T=1`, pinned to one P-core)

| Grid side | 2-D case | 3-D case | Wall clock |
|---|---|---|---|
| 100 | `lap2d100`, n=10,000 | `lap3d21`, n=9,261 | **11 s** |
| 160 | `lap2d160`, n=25,600 | `lap3d29`, n=24,389 | **3 m 38 s** |
| 200 | `lap2d200`, n=40,000 | `lap3d34`, n≈39,304 | **killed at 3 h 28 m** |
| 320 | `lap2d320`, n=102,400 | `lap3d46`, n≈97,336 | not attempted |

A 1.6x increase in grid side (100 -> 160) costs ~20x in wall clock.

## Root cause

The time is **not** in the timed solves — those are milliseconds (e.g. `chol lap3d29` solves in 14.4 ms). It is in the **untimed factorization setup** that each case performs before timing, and it is dominated by the 3-D Laplacian.

`bench_sparse.cpp:224-226` sizes the 3-D grid so its DOF count matches the 2-D one:

```cpp
const std::size_t g3 = std::max<std::size_t>(8, static_cast<std::size_t>(std::cbrt(
                           static_cast<double>(g) * static_cast<double>(g))));
```

`max(8, ...)` is only a *lower* bound; nothing caps `g3`. Equal DOF does not mean equal factorization cost — 3-D Laplacian Cholesky has far worse fill and flop growth than 2-D at the same `n`, so the 3-D case dominates and grows superlinearly. At `g=320` it is a ~97K-DOF 3-D problem.

The comment on that line says `g3` is "bounded", which describes an intent the code does not implement.

Two multipliers make it worse in the driver:

[...]

PR #344, opened 2026-08-02
docs: changelog coverage for the v5.8.0 tooling work + session log

Ravenwater:
End-of-day wrapup. Two files.

## `CHANGELOG.md`

The v5.8.0 section documented the four correctness fixes — those were added during release prep in #343 — but **none of the benchmarking, CI or documentation work that shipped in the same release**. None of #319, #320, #321, #322, #327, #329, #330 or #333 appeared anywhere in the file.

| section | added |
|---|---|
| **Added** | the benchmark results site; the benchmark-build CI lane; the Highway SIMD lane |
| **Changed** | #82 gate asserts the median with a per-size floor; `SPARSE_SIZES` default lowered to a size that completes; CI runs on draft → ready; `cmake --preset release` builds something |
| **Fixed** | `bench_klu` / `bench_superlu` did not compile; `bench_sparse`'s 3-D Laplacian grid is clamped |

These are documentation of what v5.8.0 already contains — the tag is unchanged, and the auto-generated release notes already reflected the commits.

## `docs/sessions/2026-08-02-*.md`

Follows the existing session-doc format (Objective / Context / Work Completed / Decisions & Rationale / Outcome / Issues & PRs).

The connecting thread it records: **every defect found this session lived in code that a test or lane appeared to cover but structurally could not** — n=2 and n=3 for Bunch–Kaufman where no interchange is possible, small structured matrices for SVD, plain arithmetic accumulators for the norms, "does the solve converge" for `ilu_0`, and for the tooling bugs, source files no runner compiled at all.

Three decisions worth carrying forward:

- **Fix the cause, not the instance.** Three of the four bugs shipped with a correct one-line fix in the issue, and in each the real problem was redundant or ambiguous state. `ilu_0` stored the diagonal twice with one of two consumers filtering it; storing it once makes the invariant structural. `two_norm<Acc>` could have rounded to `mag_t` and compiled, but that discards what the accumulator bought.
- **Tests must assert the failing regime is live** — that an interchange occurred, that a 2×2 block appeared, that ILU(0) was exact — otherwise they can quietly stop covering the bug, which is exactly how these survived.
- **Publish the negative results.** `ilu_0` is unchanged at 29 BiCGSTAB iterations on a 2-D Laplacian; only the tridiagonal case goes 11 → 1. And the retracted N=465 gate claim stays visible on the results page rather than being silently rewritten.

Session docs are not registered in `FILE_MAP`, matching the existing ones — they stay GitHub-only.

## Verification

`npm run build` — 43 pages, unchanged. Changelog structure intact (one `### Changed` in the 5.8.0 section, no duplicated headings).

## Summary by CodeRabbit

* **Documentation**
  * Added benchmark results and publication details for version 5.8.0.
  * Updated performance guidance, acceptance criteria, sparse benchmark defaults, and release information.
  * Documented improved numerical correctness across several linear algebra and matrix operations.

* **Bug Fixes**
  * Corrected benchmark result accessors and constrained the 3-D sparse benchmark grid for more reliable measurements.


Ravenwater:
Both findings reviewed. One fixed, one declined with reasoning.

### CSV count — valid, but the suggested resolution was the wrong way round

The arithmetic mismatch is real: the listed groups summed to 19 against a stated 20. But the **total was right and the breakdown was wrong** — the `#297` group is *six* CSVs, not five:

```
blas_sweep      5
lapack_sweep    3
gemm_scaling    4
scaling_ (#297) 6   scaling_{gemm_rect,lu,qr,chol,ewise,sparse}.csv
scoreboards     2
TOTAL          20
```

Changing the total to 19 would have made the document wrong in a way that is harder to notice than the original slip. Fixed in `0b75438` by correcting "five" to "six" and naming the six files explicitly, so the number is checkable against `benchmarks/data/` rather than needing to be trusted.

### "Full-day session" — declined

Grammatically correct as a compound modifier, but `**Duration**: Full day session` is a template field, and **all nine existing session docs use the unhyphenated form**. Changing only this one makes it the odd entry in the corpus. If the hyphenation is wanted, it should be a single sweep across all ten so the field stays uniform — which is out of scope for a wrapup PR, and I would rather not silently split the convention.

---

### Unrelated: the macOS ARM64 failure on this PR

RCA, since it is on the checks list. **Transient network failure, not a code problem.**

```
Set up sccache   ##[error]HttpError: Connect Timeout Error
Post Set up sccache   ##[error]Unable to locate executable file: undefined
```

`mozilla-actions/sccache-action` fetches the sccache binary over the network; the connect timed out 10 s in. The second error is a consequence, not an independent fault — the action never set the binary path, so its post step had nothing to run `--show-stats` against.

Four things establish it as infrastructure:

1. The job died in a **setup step, before any compilation** — 36 s total.
2. The **same lane passed on the same commit** in the sibling run (`macOS ARM64 (Apple Clang)`, 3m0s, run `30772794704`).
3. This PR is **documentation only** — `CHANGELOG.md` and one new markdown file. It cannot affect compilation on any platform.
4. All three Linux ARM64 lanes passed; ARM is not implicated.

Not recurrent either: the other two recent red runs are on `feat/ryzen-9-8945hs-benchmarks` with a genuine step failure, a different cause.

Re-running the failed job rather than pushing a no-op commit.

Issue #110, opened 2026-06-03
perf(operation): decouple GEMM parallel granularity from the mc cache block


Ravenwater:
All parallelism in the native blocked GEMM comes from the **ic** loop, so the number of parallel work units is exactly `nib = ceil(m / mc)`. That makes **`mc` both the L2 cache-blocking parameter and the unit of parallel decomposition**, with no independent control — growing it for cache reasons directly starves the thread decomposition.

This is what now limits parallel efficiency at the physical-core count. It is *not* the original diagnosis of this ticket (persistent pool + multi-loop parallelism); that work has since landed — see [History](#history) below.

## Current state (i7-12700K, fp64, N=2048, 8 P-cores, pinned)

| T | GFLOP/s |
|---:|---:|
| 1 | 57.83 |
| 2 | 112.56 |
| 4 | 209.28 |
| 8 | 367.51 |

**6.36× / 79% efficiency**, against OpenBLAS 7.14% / 89% and BLIS 7.39× — BLIS scales best, which is notable since it is the multi-loop design being emulated.

## Why the ic loop is the only source of parallelism

Real blocking parameters for the `native-fast` build (Highway, SIMD width 4):

```
mr=4  nr=8  kc=256  mc=64  nc=4096      packA = 0.13 MB   packB = 8.39 MB
```

`nc = 4096`, so the whole n dimension fits in **one** jc block for every square size up to 4096:

| N | ic-blocks | jc-blocks | grid at T=2 / 4 / 8 |
|---|---:|---:|---|
| 512 | 8 | **1** | 2×1, 4×1, 8×1 |
| 1024 | 16 | **1** | 2×1, 4×1, 8×1 |
| 2048 | 32 | **1** | 2×1, 4×1, 8×1 |
| 4096 | 64 | **1** | 2×1, 4×1, 8×1 |
| 8192 | 128 | 2 | 2×1, 4×1, 8×1 |

The `ic_nt × jc_nt` grid machinery exists and is correct, but jc parallelism is **structurally unavailable** at these sizes. (At N=8192, where `njb = 2`, the fill-ic-first policy still takes the whole budget for ic — a secondary issue.)

## The measurement that identifies the coupling

`hw_traits` describes a Haswell core (256 KB L2) while the i7-12700K P-core has 1.25 MB. Correcting only the L2 term moves `mc` from 64 to 320 — and makes performance **worse**:

| | T=1 | T=2 | T=4 | T=8 | speedup | efficiency |
[...]

PR #350, opened 2026-08-03
docs(benchmarks): refresh the i7-12700K threaded measurements after #348

Ravenwater:
Relates to #110. Follow-up to #348, which left the committed scaling data and result page stale.

## Refreshed: GEMM scaling (N=2048, all four backends re-run together)

| T | native-fast | blis | openblas | mkl |
|---:|---|---|---|---|
| 1 | 55.5 (1.00×) | 71.1 (1.00×) | 73.4 (1.00×) | 75.0 (1.00×) |
| 2 | 111.6 (2.01×) | 141.6 (1.99×) | 146.4 (1.99×) | 147.0 (1.96×) |
| 4 | 216.1 (3.90×) | 276.8 (3.89×) | 287.2 (3.91×) | 288.1 (3.84×) |
| 8 | **401.8 (7.24×)** | 520.9 (7.33×) | 525.5 (7.16×) | 539.4 (7.19×) |

**native-fast now scales better than OpenBLAS and MKL** — 7.24× / 90.5% efficiency against 7.16× / 89.5% and 7.19× / 89.9%. Only BLIS leads, at 7.33×.

All four were re-run in the same pass so they stay comparable, even though only native-fast changed.

**The headline is the ratio, not the speedup.** native-fast throughput relative to OpenBLAS is **75.5% at one thread and 76.4% at eight** — essentially identical, where it previously degraded from ~80% to ~62% as threads were added. Threading is no longer what separates them; the remainder is the single-core kernel gap (#82).

## Refreshed: #297 kernel families

`gemm_rect` and `ewise` route through the same blocked GEMM and moved with it:

| family | before | after |
|---|---|---|
| `gemm_rect` | 6.43× | **7.07×** |
| `ewise` | 5.13× | **7.12×** |

`lu`, `qr`, `chol` and `sparse` moved only within noise.

## Deliberately not re-run

The single-threaded `blas_` / `lapack_` sweeps and the sparse scoreboards. #348 touches only the parallel region, and the serial path measured 57.83 → 57.30 GFLOP/s — noise. Re-running would churn the committed CSVs without adding information.

That makes the page's original claim — *"Measured 2026-08-01 in a single session: every figure below comes from the same set of builds"* — no longer true, so it is replaced with a **Provenance** paragraph stating which sections come from which date and why the rest were left alone. Quietly leaving the old sentence in place would have been the easy option and the wrong one.

I also corrected a sentence asserting per-thread sparse wall clocks (230 s, 230 s, 227 s) that **this** pass did not re-time; it is now attributed to the run that did measure them, rather than presented as current.

## Layout

Figures regenerated into `docs/img/benchmarks/intel-i7-12700k/` per the per-system layout introduced in #345, and `benchmarks/data/gemm_scaling.png` (referenced by `data/README.md`) refreshed alongside. The orphaned top-level figures from the pre-#345 layout are already gone.

## Verification

`npm run build` — 44 pages, all four i7 figures resolving to `/mtl5/img/benchmarks/intel-i7-12700k/…`, refreshed numbers present in the built HTML.

## Summary by CodeRabbit

* **Documentation**
  * Updated benchmark results to distinguish single-threaded and threaded measurements.
  * Revised multi-core GEMM, rectangular, elementwise, and sparse-operation performance figures.
  * Updated conclusions on scaling behavior and comparisons with OpenBLAS.


Ravenwater:
Both findings addressed in `06568ba`. Both were correct, and the first one caught a genuine error rather than a wording issue.

### 1. `ewise` attribution — wrong, and the correction matters

I wrote that `gemm_rect` and `ewise` both improved "with the parallel B pack (#348) — both route through the same blocked GEMM". That is false for `ewise`. Checking `bench_ewise`:

```cpp
// Vectors: y = a + b ...
auto t = measure([&]{ c = a + b; sink += c(n - 1); }, ...);
// Matrices: C = A + B (row-parallel sweep)
```

Pure element-wise addition through `detail::parallel_ewise`, with **zero** references to `mult` or `gemm` anywhere in the function. #348 cannot have caused its 5.13× → 7.12× move.

`gemm_rect`'s improvement stands — it genuinely does route through the blocked GEMM. The `ewise` change is now recorded as **unattributed**: these are bandwidth-bound measurements and the honest statement is "treat as run-to-run variation until something isolates it", not a made-up mechanism. Inventing a plausible cause for a number I had not traced is exactly the failure this review caught.

### 2. Provenance vs the committed data

Also correct. The note offered *"57.83 → 57.30 GFLOP/s at one thread"* as evidence the serial path is unchanged, but those came from the **#348 verification runs in a different session** — the committed sweep records **55.46** for the same configuration (N=2048, T=1, one P-core). Quoting unsourced numbers next to a table that says 55.5 is confusing.

The argument now rests on the code fact — the serial path bypasses the threaded region entirely, so #348 cannot touch it — and the three `T = 1` figures (57.83, 57.30, 55.46) are presented together as a **bound on session-to-session variation (~4%)** rather than as a before/after. That is more useful anyway: it tells a reader what spread to discount.

### The same problem in `data/README.md`

Following the metadata thread found it there too. `Run date | 2026-08-01 (all files below, one session)` stopped being true the moment the threaded CSVs were re-measured on 08-03. It now states which files come from which date, and notes that the four GEMM-scaling backends remain comparable **with each other** because they were re-run together — which is the property that actually matters for that table.

`npm run build`: 44 pages, unchanged.

Issue #351, opened 2026-08-03
perf(simd): single-core GEMM sustains ~72-78% of FMA peak — identify the microkernel limiter


Ravenwater:
Split out of the observation in #110 / #348 that, with parallel efficiency now at ~90%, **threading is no longer what separates native-fast from the tuned libraries** — the remaining difference is entirely single-core.

Not a reopening of #82. That epic's target was "within 10–20% of hand-optimized OpenBLAS/MKL", and the current median is **82.3% of OpenBLAS for N ≥ 256** — 17.7% off, inside the band. It closed having delivered that. This issue is the narrower, still-unanswered question: *why* the microkernel sustains ~three-quarters of peak rather than the ~90% the vendors reach.

## Measured (i7-12700K, fp64, one P-core, FMA peak taken as ~78 GFLOP/s)

| N | native-fast | openblas | % openblas | % FMA peak |
|---:|---:|---:|---:|---:|
| 145 | 52.95 | 66.62 | 79.5% | 67.9% |
| 305 | 56.23 | 69.72 | 80.6% | 72.1% |
| 545 | 56.89 | 70.92 | 80.2% | 72.9% |
| 785 | 60.99 | 72.12 | 84.6% | 78.2% |
| 1025 | 59.44 | 72.37 | 82.1% | 76.2% |
| 2048 | 55.5 | 73.4 | 75.5% | 71.2% |

So **72–78% of peak**, against roughly 89–94% for OpenBLAS, BLIS and MKL on the same machine and sizes. The gap widens at large N: 75.5% of OpenBLAS at N=2048 versus the 82.3% median over N ≥ 256.

## One explanation is already ruled out

It is **not** cache blocking derived for the wrong CPU. `hw_traits` does describe a Haswell part (256 KB L2) on a machine with 1.25 MB, but correcting only the L2 term moves `mc` 64 → 320 and makes single-thread performance **worse — 57.83 → 51.12 GFLOP/s, −12%**. Bigger blocks are not the answer here. (Measurement in #110; the related `nc`/L3 derivation problem is #347.)

## Leading hypothesis: the register tile is derived at exactly the minimum

`simd::derive_blocking` sizes the microtile as

```cpp
const std::size_t area = nvec * hw.fma_latency * hw.fma_units;   // 4 * 4 * 2 = 32
std::size_t nr = round_up(isqrt_ceil(area), nvec);               // 8
std::size_t mr = ceil_div(area, nr);                             // 4
```

[...]

PR #354, opened 2026-08-03
feat(ppe): Platform Performance Engineering lab — GEMM optimization progression

Ravenwater:
A new top-level `ppe/` directory: a laboratory for measuring where GEMM performance comes from and where it goes, so the lessons from the recent performance work are captured somewhere reusable instead of living in issue comments.

**Draft** — the harness, kernels and write-ups are complete and the measurement sweep is running now. Data, figures and `docs/results.md` land in a second commit.

## The progression

Six kernels, each adding exactly **one** idea to the previous, so a measured delta is attributable to that idea:

| step | idea added |
|---|---|
| `v0_naive` | none — textbook `ijk`, also the correctness reference |
| `v1_ikj` | loop reorder for a unit-stride inner loop |
| `v2_blocked` | cache blocking (`MC`/`NC`/`KC`) |
| `v3_packed` | pack tiles into contiguous panels |
| `v4_regtile` | accumulate an `MR × NR` microtile in locals |
| `v5_micro` | compile-time `MR`/`NR`, no edge tests, fully unrollable |

Measured across **seven operand types on one throughput axis** (GOP/s, since half the matrix is integer): `int8`/`int16`/`int32`/`int64` with widening accumulators as real integer GEMM does, and `fp16`/`fp32`/`fp64`.

**Every kernel is verified against `v0`** for the same type and size before its timing is recorded — exactly for integers, tight relative tolerance for floats. A fast wrong answer cannot become a result.

Standalone by design: nothing includes `<mtl/...>`, so each step is isolated.

## Headline finding, already reproduced

`v5` — the "fully optimized" step — is **slower than a plain loop reorder** for several types (7.3 vs 37.3 GOP/s for fp32 at N=512). Not a bug; the compiler says why:

```
missed: not vectorized: loop nest containing two or more
        consecutive inner loops cannot be vectorized
stack usage:  v4_regtile 832 B  ->  v5_micro 1344 B
```

Full unrolling of the 4×8 tile replaces a vectorizable inner loop with 32 *scalar* accumulators, which do not fit in 16 registers and spill. **Unrolling and vectorization compete for the same registers**, and source-level "optimization" can silently flip which one you get.

That is the same mechanism as **#351** (MTL5's own microkernel derives `mr × nr = 4 × 8`, exactly the minimum accumulator count, and sustains ~72–78% of peak). PPE is the cheap place to test it — a one-parameter tile sweep with no library API to preserve.

## On measuring "efficiency"

Two approaches were discarded, and the failures are documented in `include/ppe/peak.hpp` because they are instructive:

1. **An ISA model that put `int64` at 2 ops/cycle was exceeded — 119%.** AVX2 has no 64-bit SIMD multiply, but the compiler emulates one, so the real ceiling is higher than "scalar". *A model that can be exceeded is not a ceiling.* The driver now fails with `PEAK MODEL TOO LOW` if any result exceeds it — which is what caught this.
2. **A probe that *measured* achievable rate** reported **3.8 million GOP/s** with literal operands (what a deleted loop looks like), and after making them opaque reported **17.7 GOP/s for fp64 — below the ~37 the GEMM kernels actually reach**, because the accumulators did not stay in vector registers. *A probe beaten by what it bounds is not a ceiling either.*

The model that ships is ISA-anchored, written out in full, and validated on every run.

## Write-ups

Each step has a document stating the change, the mechanism, what to watch in the results, and — where relevant — hypotheses with the experiments that would test them. `docs/05-vectorization-cliff.md` carries five numbered hypotheses (H1–H5) for the regression above, with the two cheapest flagged to run first.

## Test plan

- [ ] Measurement sweep completes; data + figures + `docs/results.md` pushed
- [ ] Promote to ready: `gh pr ready <N>`


Ravenwater:
## Measurements in — 252 results, all verified, and the tidy story does not survive

Sweep complete and pushed in `d6f628c`. Every one of the 252 measurements verified against the naive reference; none exceeded the peak model.

### Four different kernels win across seven types

| type | peak | v0 | v1 | v2 | v3 | v4 | v5 | best | vs v0 |
|---|---:|---:|---:|---:|---:|---:|---:|---|---:|
| int8 | 160 | 1.4 | 34.5 | 32.8 | 26.9 | 19.2 | **36.4** | v5 | 26× |
| int16 | 160 | 1.0 | **48.1** | 43.8 | 36.0 | 13.9 | 35.4 | v1 | 48× |
| int32 | 80 | 1.0 | 23.6 | 30.7 | 35.4 | 20.5 | **35.9** | v5 | 37× |
| int64 | 40 | 0.6 | 5.7 | 11.5 | **12.9** | 11.7 | 9.6 | v3 | 20× |
| fp16 | 160 | 1.0 | 5.4 | 5.0 | **29.9** | 20.9 | 7.3 | v3 | 30× |
| fp32 | 160 | 1.0 | 20.8 | 21.3 | **28.6** | 21.0 | 7.3 | v3 | 28× |
| fp64 | 80 | 0.6 | 6.3 | 12.5 | 17.2 | **23.8** | 8.3 | v4 | 37× |

GOP/s at N=2048. The expectation going in was a monotone progression; four different winners is the most useful thing the study produced.

### The cliff turned out to be something other than what I wrote

The draft attributed `v5`'s regression to register pressure and spills. The full sweep contradicted that: `v5` is the **best** kernel for `int8` and `int32`, and catastrophic only for the floats. Checking the compiler per type:

```
v5 micro <int32 > : 0 vectorized loops
v5 micro <double> : 0 vectorized loops
v5 micro <float > : 0 vectorized loops
```

**`v5` never vectorizes, for any type.** The spill is secondary. The real mechanism:

> the compile-time-unrolled microkernel is always scalar, and the penalty for being scalar is proportional to how much that type gains from vectorization

`fp32`/`fp64` gain 4–8× from SIMD, so losing it is fatal. `int32` is throughput-limited by `vpmulld` rather than lane-limited, so scalar is competitive and `v5`'s tighter scheduling wins on the margin.

That **answers H5 and demotes H1** — register pressure is not first-order — so H2 (keep the tile, keep the loop) is now the experiment to run first. The write-up was corrected rather than left as originally drafted.

### Other findings

- **`v0 → v1`, a one-line loop reorder, is the biggest step in the whole progression** at 20–48×. Any performance work that starts with blocking is optimizing the wrong thing.
- **Cache blocking is small or negative** at these sizes — `int16` *loses* 9%.
- **Packing is where the floats come alive** (`fp16` 5.0 → 29.9) and where narrow integers regress, which the widening pack in `v3` explains; H6 would confirm it.
- **Nothing gets close to peak** (17.9%–44.9%), as expected without explicit SIMD, prefetch or hand scheduling. The informative number is the spread: `fp32` at 17.9% has far more headroom than `int32` at 44.9%, precisely because it has more to gain from vectorization these kernels do not exploit.

### On #351

This does **not** show MTL5's `4 × 8` tile is wrong — MTL5 *does* vectorize its microkernel, so it is not on this cliff. What it shows is that the margin between "vectorized with a minimal tile" and "unrolled into scalars" is **one compiler decision wide**, and a tile derived at exactly the saturation minimum has no slack if that decision flips.

Ready to promote whenever you are.

PR #358, opened 2026-08-03
fix(mat): reject col_major compressed2D at compile time

### LINKED ISSUE #355: bug(mat): compressed2D accepts tag::col_major but ignores it — the container is always row-major
`compressed2D` takes a `Parameters` bundle whose `orientation` may be `tag::row_major` or `tag::col_major` (`mat/parameter.hpp:14`), but nothing in the container ever reads it. A `col_major` `compressed2D` is byte-for-byte a CSR matrix, and every operation treats it as one.

The parameter is accepted silently, so a caller who asks for CSC gets CSR without any diagnostic — and if they populate it from real CSC arrays, they get the transpose while the constructor reports success.

`dense2D` gets this right (`dense2D.hpp:35,60,67` branch on `typename Parameters::orientation` with `if constexpr`), which is what makes the `compressed2D` behaviour look like an oversight rather than an intentional restriction. `compressed2D.hpp` contains **zero** occurrences of `orientation`, `row_major` or `col_major`.

Found at `c666ae4` while binding the alternative sparse storage formats for Python.

## Reproducer

```cpp
#include <cstdio>
#include <mtl/mat/compressed2D.hpp>
#include <mtl/mat/inserter.hpp>
#include <mtl/vec/dense_vector.hpp>
#include <mtl/operation/mult.hpp>

using CSR = mtl::mat::compressed2D<double>;
using CSC = mtl::mat::compressed2D<double,
            mtl::mat::parameters<mtl::tag::col_major>>;
using Vec = mtl::vec::dense_vector<double>;

// A (2x3) = [[1,2,0],
//            [0,3,4]]
template <class M> void build(M& A) {
    mtl::mat::inserter<M> ins(A);
    ins[0][0] << 1.0; ins[0][1] << 2.0; ins[1][1] << 3.0; ins[1][2] << 4.0;
}

template <class M> void report(const char* label, M& A, std::size_t want_major) {
    build(A);
    std::printf("%-4s major.size()=%zu (want %zu)  major:", label,
                A.ref_major().size(), want_major);
    for (auto v : A.ref_major()) std::printf(" %zu", (std::size_t)v);
    std::printf("  minor:");
    for (auto v : A.ref_minor()) std::printf(" %zu", (std::size_t)v);
    std::printf("\n");
}

int main() {
    CSR a(2,3); report("CSR", a, 3);   // nrows + 1
    CSC b(2,3); report("CSC", b, 4);   // ncols + 1
    std::printf("\nidentical internal arrays: %s\n",
        (a.ref_major()==b.ref_major() && a.ref_minor()==b.ref_minor()
         && a.ref_data()==b.ref_data()) ? "YES -- the tag changed nothing" : "no");

    // Feed a col_major matrix genuine scipy CSC arrays for
    // B = [[1,2,0],[0,3,4],[5,0,6]]:
    //   indptr=[0,2,4,6] indices=[0,2,0,1,1,2] data=[1,5,2,3,4,6]
    std::size_t indptr[4]  = {0,2,4,6};
    std::size_t indices[6] = {0,2,0,1,1,2};
    double      data[6]    = {1,5,2,3,4,6};
    CSC B(3,3,6,indptr,indices,data);
    Vec x(3), y(3); x[0]=1; x[1]=2; x[2]=3;
    mtl::mult(B, x, y);
    std::printf("\nB@x = [%.0f %.0f %.0f]   B@x should be [5 18 23]"
                "   (B^T@x is [16 8 26])\n", y[0], y[1], y[2]);
    std::printf("B(2,0)=%.0f (want 5)   B(0,2)=%.0f (want 0)\n", B(2,0), B(0,2));
}
```

```
CSR  major.size()=3 (want 3)  major: 0 2 4  minor: 0 1 1 2
CSC  major.size()=3 (want 4)  major: 0 2 4  minor: 0 1 1 2

identical internal arrays: YES -- the tag changed nothing

B@x = [16 8 26]   B@x should be [5 18 23]   (B^T@x is [16 8 26])
B(2,0)=0 (want 5)   B(0,2)=5 (want 0)
```

Three separate symptoms, all from the same cause:

1. **The major array is the wrong length.** For a 2×3 matrix CSC needs `ncols + 1 = 4` starts; it builds 3.
2. **The tag is inert.** A `row_major` and a `col_major` matrix built from the same elements have identical `major`, `minor` and `data`.
3. **Real CSC input silently becomes the transpose.** `mult` returns `B^T @ x`, and element access agrees — `B(0,2)` returns the value that belongs at `B(2,0)`.

## Where it comes from

`operator()` indexes the majors by row unconditionally:

```cpp
// compressed2D.hpp:53
value_type operator()(size_type r, size_type c) const {
    auto begin = indices_.begin() + starts_[r];
    auto end   = indices_.begin() + starts_[r + 1];
```

and the array constructor's own assert assumes the same:

```cpp
// compressed2D.hpp:46
assert(starts_[nrows_] == nnz_count);
```

The inserter and `mult` follow suit, which is why everything stays self-consistently row-major and the wrongness only shows when the data genuinely came from a column-major source.

## Suggested resolution

1. **Smallest:** `static_assert` that `orientation` is `tag::row_major` in `compressed2D`. Turns a silently wrong matrix into a compile error, and costs nothing for existing callers since none can be relying on the current behaviour to be correct.
2. **Complete:** implement the orientation the way `dense2D` already does — branch on `typename Parameters::orientation` for the major/minor mapping in the constructor, `operator()`, the inserter and the `mult` kernels. This is what would give MTL5 a real CSC, and with it a zero-copy path from `scipy.sparse.csc_matrix`.

[mtl5-python#30](https://github.com/stillwater-sc/mtl5-python/pull/30) binds COO and ELL but deliberately not CSC, and converts `csc_matrix` input to CSR at the boundary. The transpose-SpMV that CSC would have enabled is provided instead by `mtl::trans`, which is correct for sparse.

Ravenwater:
## Summary

`compressed2D` accepted `tag::col_major` and ignored it. The container is CSR unconditionally — `starts_` is indexed by row in `operator()`, in the array constructor's own assert, in the inserter and in every `mult` kernel — but nothing ever read `Parameters::orientation`.

So a `col_major` instantiation was byte-for-byte a CSR matrix. **That is worse than unsupported**: a caller who populated it from genuine CSC arrays got the transpose while the constructor reported success — `mult` returned `A^T*x`, `A(0,2)` returned the value belonging at `A(2,0)`, no diagnostic anywhere (#355).

It is now a compile error.

## Which resolution, and why

The issue offers two. I took the **smallest** (`static_assert`), not the complete one (implement CSC), and the reason is worth stating rather than assuming:

**34 headers across `sparse/`, `itl/`, `interface/`, `io/`, `mat/` and `operation/` read the raw arrays as CSR.** Real CSC support means generalizing the major/minor mapping in every one of them, and each is a place a silent wrong answer could be reintroduced — the same defect this PR is closing. That is a feature, and a large one, not a bug fix. mtl5-python already works around it by converting `csc_matrix` at the boundary and using `mtl::trans` for transpose-SpMV, which is correct for sparse.

The guard costs existing callers nothing: I checked, and **nothing in the tree instantiates a `col_major` `compressed2D`.** None could have — no such matrix was ever correct.

## The compile-failure test, and why it is not `WILL_FAIL`

A guard nobody tests is a guard that gets deleted. New harness at `tests/unit/compile_fail/`: sources that must not compile, built on demand (`EXCLUDE_FROM_ALL`) so normal builds are unaffected.

It uses `PASS_REGULAR_EXPRESSION` against an `// EXPECT-ERROR: <regex>` line in each source, **not `WILL_FAIL`**. `WILL_FAIL` passes whenever the build returns non-zero, so a typo in the test source — a misspelled header, a missing semicolon — would make it pass green while proving nothing. Matching the expected diagnostic means it only passes when compilation fails for the *right* reason. CMake errors out at configure time if a compile-fail source has no `EXPECT-ERROR` line.

**Negative control, run before committing:** removing the `static_assert` turns the test red; restoring it turns it green. It can actually fail.

## Changes

| file | what |
|---|---|
| `include/mtl/mat/compressed2D.hpp` | `static_assert` on `Parameters::orientation`, with the reasoning inline |
| `tests/unit/compile_fail/compressed2d_col_major.cpp` | must-not-compile case |
| `tests/unit/CMakeLists.txt` | compile-failure harness |
| `tests/unit/mat/test_compressed2D.cpp` | two runtime regression cases |
| `docs/sparse-direct-solvers-design.md` | corrects a stale claim (below) |

The runtime cases pin the half of the contract a compile-failure test cannot: that the row-major layout the container implements is the one it documents. The major array has length `nrows+1` — the issue's 2×3 case, where CSC would need 4 starts and got 3 — and raw arrays read as CSR give `A*x` rather than `A^T*x`, with `trans()` supplying the transposed product correctly.

## Docs correction found along the way

`docs/sparse-direct-solvers-design.md` advertised a *"Zero-cost CSC view over `compressed2D` with col-major parameters"*. **That was never implemented** — `sparse/util/csc.hpp` uses a standalone owning `csc_matrix` struct — and it pointed readers at exactly the construct this PR makes a compile error. Corrected, along with the file-tree comment above it. (`docs/design/capability-assessment-and-expansion.md` was already accurate: *"CSC exists only as a conversion utility"*.)

## Noted, deliberately not fixed here

The two sibling sparse containers have the same inert-orientation hole, and they are **not** equally severe:

| container | reads `orientation`? | severity |
|---|---|---|
| `compressed2D` | no | **wrong answers** — fixed here |
| `ell_matrix` | no | same defect class: `indices_[r*width_+k]` is row-major padded, so a `col_major` instance is silently row-major ELL |
| `coordinate2D` | no | benign — stores explicit `(row,col,val)` triplets, so orientation cannot change the result; only `sort()` ordering is conventional |

Left out of scope rather than folded in silently. `ell_matrix` wants its own issue; the harness added here makes the same guard a few lines when someone takes it.

## Test results

| | build | tests |
|---|---|---|
| gcc | clean | 157/157 |
| clang | clean | 157/157 |

Resolves #355

### DIFF
--- a/docs/architecture/aggregate-types.md
+++ b/docs/architecture/aggregate-types.md
@@ -188,7 +188,7 @@ MTL5's ndarray design is informed by [xtensor](https://github.com/xtensor-stack/
 | Dynamic rank | `xt::xarray<T>` | `xarray<T>` |
 | Custom scalars | Limited | First-class (posit, lns, cfloat, ...) |
 | Linear algebra | Separate (xtensor-blas) | Integrated (MTL5 operations + BLAS dispatch) |
-| Sparse matrices | No | Full support (CSR, CSC, COO, ELLPACK) |
+| Sparse matrices | No | CSR, COO, ELLPACK (CSC only as a conversion utility) |
 | Iterative solvers | No | Integrated (CG, GMRES, BiCGSTAB, ...) |
 | Tensor algebra | No | Separate `tensor<>` with Einstein notation |
 | Expression templates | Yes (lazy) | Yes (lazy, compatible with MTL5 expressions) |
--- a/docs/sparse-direct-solvers-design.md
+++ b/docs/sparse-direct-solvers-design.md
@@ -74,7 +74,7 @@ include/mtl/
 │   │   ├── factor_result.hpp        # Factorization result types (symbolic + numeric)
 │   │   └── triangular_solve.hpp     # Sparse triangular solve (reach + solve)
 │   ├── util/
-│   │   ├── csc.hpp                  # CSC view/adapter for compressed2D
+│   │   ├── csc.hpp                  # Owning csc_matrix + CRS<->CSC conversion
 │   │   ├── permutation.hpp          # Permutation vector operations
 │   │   ├── dulmage_mendelsohn.hpp   # BTF decomposition
 │   │   └── scatter.hpp              # Sparse accumulator (scatter/gather)
@@ -207,7 +207,7 @@ Build the reusable infrastructure that all three factorizations depend on.
 
 | Component | File | Description |
 |-----------|------|-------------|
-| CSC adapter | `sparse/util/csc.hpp` | Zero-cost CSC view over `compressed2D` with col-major parameters; CRS-to-CSC conversion |
+| CSC adapter | `sparse/util/csc.hpp` | Standalone owning `csc_matrix` struct plus CRS-to-CSC / CSC-to-CRS conversion |
 | Permutation | `sparse/util/permutation.hpp` | Apply/invert/compose permutation vectors; permuted matrix views |
 | Scatter | `sparse/util/scatter.hpp` | Sparse accumulator for column assembly (the workhorse of sparse arithmetic) |
 | Elimination tree | `sparse/analysis/elimination_tree.hpp` | O(nnz) etree construction from CSC |
--- a/include/mtl/mat/compressed2D.hpp
+++ b/include/mtl/mat/compressed2D.hpp
@@ -1,12 +1,15 @@
 #pragma once
 // MTL5 -- CRS (Compressed Row Storage) sparse matrix
-// Row-major only for Phase 4. Three-array storage: data_, indices_, starts_.
+// Row-major ONLY, and now enforced -- see the static_assert below (#355).
+// Three-array storage: data_, indices_, starts_.
 #include <algorithm>
 #include <cassert>
 #include <cstddef>
+#include <type_traits>
 #include <vector>
 
 #include <mtl/config.hpp>
+#include <mtl/tag/orientation.hpp>
 #include <mtl/tag/sparsity.hpp>
 #include <mtl/traits/category.hpp>
 #include <mtl/traits/ashape.hpp>
@@ -25,6 +28,25 @@ public:
     using size_type       = typename Parameters::size_type;
     using const_reference = const Value&;
     using reference       = Value&;
+
+    // compressed2D is CSR, unconditionally: `starts_` is indexed by ROW
+    // everywhere -- operator(), the array constructor's own assert, the
+    // inserter and every mult kernel. Nothing in this container has ever read
+    // Parameters::orientation.
+    //
+    // A col_major instantiation was therefore accepted silently and was
+    // byte-for-byte a CSR matrix. That is worse than unsupported: a caller who
+    // populated it from genuine CSC arrays got the TRANSPOSE while the
+    // constructor reported success -- mult returned A^T*x, and A(0,2) returned
+    // the value belonging at A(2,0), with no diagnostic anywhere (#355).
+    //
+    // Reject it at compile time instead. This costs existing callers nothing:
+    // no col_major compressed2D can have been correct, so none can exist.
+    static_assert(std::is_same_v<typename Parameters::orientation, tag::row_major>,
+        "compressed2D is CSR (row-major) storage only; a col_major instantiation "
+        "would be byte-identical to CSR and would silently read CSC input as its "
+        "transpose (#355). Convert CSC to CSR at the boundary, and use mtl::trans "
+        "for a transposed sparse product.");
 
     // -- Constructors ----------------------------------------------------
 
--- a/include/mtl/mat/ell_matrix.hpp
+++ b/include/mtl/mat/ell_matrix.hpp
@@ -1,11 +1,14 @@
 #pragma once
 // MTL5 -- ELLPACK sparse matrix format
-// Fixed-width per-row storage. Good for GPU and matrices with uniform row lengths.
+// Fixed-width per-ROW storage, and now enforced -- see the static_assert below
+// (#355). Good for GPU and matrices with uniform row lengths.
 #include <algorithm>
 #include <cassert>
 #include <cstddef>
+#include <type_traits>
 #include <vector>
 
+#include <mtl/tag/orientation.hpp>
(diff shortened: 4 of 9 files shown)


Ravenwater:
## Folded in: `ell_matrix` now rejects `col_major` too (1735b00)

The sibling container had the same inert-orientation defect. **The severity is different, and I want to state it precisely rather than by analogy**, because "same bug as compressed2D" would overstate it.

`ell_matrix` is row-padded unconditionally — both arrays are `nrows*width`, every access is `indices_[r * width_ + k]` — and `Parameters::orientation` was never read. But the **only** fill path is the `compressed2D` constructor, which for `col_major` parameters became a compile error in the previous commit. So no `col_major` ELL can be populated, and none can be returning wrong answers.

What it *could* still do — verified, not assumed:

```
col_major ell_matrix constructed: rows=2 cols=3 width=2
indices array size=4 -> nrows*width=4 (row-major padding)
```

It constructs, empty, while claiming a layout it does not implement. The `compressed2D` guard does not reach it: **declaring a reference parameter of type `compressed2D<Value, Parameters>` does not instantiate that class**, so the `static_assert` inside never fires.

So this commit is not a wrong-answer fix. It closes a hole that reopens silently the moment an inserter or setter is added, and replaces a confusing diagnostic about `compressed2D` with one at the point of misuse.

**Negative control:** removing this `static_assert` fails *only* the new compile-failure test and leaves the `compressed2D` one green — the two are independently discriminating, not coupled.

```
1/2 compile_fail_compressed2d_col_major ... Passed
2/2 compile_fail_ell_matrix_col_major ..... ***Failed  Required regular expression not found.
```

### `coordinate2D` is deliberately still untouched

It stores explicit `(row, col, value)` triplets, so orientation genuinely cannot change a result — `insert(r,c,v)` and `operator()(r,c)` name both indices. Only `sort()` ordering is conventional, and no caller can be misled about layout. Adding a guard there would be noise, not safety.

### Second docs overclaim found

`docs/architecture/aggregate-types.md`'s xtensor comparison table claimed MTL5 has *"Full support (CSR, CSC, COO, ELLPACK)"*. CSC is not a first-class matrix type — it exists only as the conversion utility in `sparse/util/csc.hpp`. That is the second doc in this PR promising a CSC that does not exist; both were pointing readers toward the construct these commits make a compile error.

Also exposes `param_type` on `ell_matrix`, as `compressed2D` already does, so orientation is assertable from outside the class.

**gcc and clang: full build clean, 158/158.**

Issue #381, opened 2026-08-04
bug(test): three MT sparse solve tests fail under -march=native — bit-exact comparison across two different loop nests


Ravenwater:
Found while sweeping blocking parameters for #351, in a Release + Highway + `MTL5_NATIVE_ARCH=ON` build.

## Symptom

```
sparse_test_supernodal_ldlt_mt          FAILED
sparse_test_supernodal_lu_mt            FAILED
sparse_test_trisolve_level_schedule_mt  FAILED
```

with the giveaway expansion:

```
tests/unit/sparse/test_supernodal_ldlt_mt.cpp:118: FAILED:
  REQUIRE( x(static_cast<int>(i)) == ref[i] )
with expansion:
  6.3137805202 == 6.3137805202
```

Identical to print precision, differing in the last bits.

## Isolated to `-march=native`

Same build type, same Highway, one flag apart:

| configuration | result |
|---|---|
| Release + Highway + `MTL5_NATIVE_ARCH=ON` | **FAIL** |
| Release + Highway + `MTL5_NATIVE_ARCH=OFF` | pass — 5704 assertions |
| Debug, no Highway | pass |

Not thread oversubscription: they fail when run serially under `ctest -R`, one at a time.

## Why

The tests assert **bit-exact** equality between the threaded solve and a reference, and the file header explains the reasoning:

> Reference solve using the serial dense unit kernels on the **SAME factor**
> ... steps are unchanged serial loops, so exact equality (`==`) is the correct bar.

The *intent* is right — same arithmetic, same order, serial versus threaded. But the reference is a **separate loop nest in separate source**, not the same code compiled twice. Under `-march=native` the compiler is free to contract one into FMAs and not the other, and it does. The mathematics is identical; the rounding is not.

This is worth contrasting with a case where `==` **is** the right bar. In #356, `test_ldlt.cpp` asserts `ldlt_h_solve == ldlt_solve` bit-exactly for real input, and that survives `-march=native` because the two function bodies are the *same expression tree* — `conj<T>::apply` inlines to the identity, so any contraction the compiler applies, it applies to both. The distinction is not "exact vs tolerance", it is **same expression tree vs different loop nests**. These three tests are on the wrong side of that line.

## Why CI does not see it

The Highway lane deliberately omits the flag, with a documented rationale (`.github/workflows/ci.yml`):

[...]

Issue #347, opened 2026-08-03
perf(simd): nc derivation sizes the packed-B panel to 100% of L3 and ignores that L3 is shared


Ravenwater:
Split out of #110, which needs this fixed before its option 1 (shrink `nc` so the jc loop becomes usable) is viable.

## The derivation targets 100% of L3, at every L3 size

`simd::derive_blocking` sizes `nc` so the packed-B panel exactly fills the assumed L3. Varying only `l3_bytes` (Highway build, `double`, SIMD width 4):

| l3 assumed | kc | nc | packB | packB / L3 | jc-blocks at N=2048 |
|---|---:|---:|---:|---:|---:|
| 8 MB (current default) | 256 | 4096 | 8.39 MB | **1.00** | 1 |
| 16 MB | 256 | 8192 | 16.78 MB | **1.00** | 1 |
| 25 MB (real i7-12700K) | 256 | 12800 | 26.21 MB | **1.00** | 1 |
| 32 MB | 256 | 16384 | 33.55 MB | **1.00** | 1 |
| 64 MB | 256 | 32768 | 67.11 MB | **1.00** | 1 |

The ratio is invariant: `nc` scales linearly with the assumed L3 so that `packB ≈ L3` always.

## Three problems with that

**1. No headroom.** A panel sized to occupy the entire L3 leaves nothing for the packed A panel, the C macro-block, or streaming traffic — all of which move through the same cache concurrently.

**2. L3 is shared, and the derivation does not know it.** `l3_bytes` is documented as "shared L3 (per core group)", but `nc` is chosen as though one consumer owns it. With `jc_nt` teams each holding its own packed-B, the aggregate demand is `jc_nt × L3`. This is currently masked because `jc_nt` is always 1 (#110) — but making jc parallelism usable is exactly what #110 proposes, so this would bite immediately.

**3. Correcting the stale L3 value makes things worse, not better.** The default describes a Haswell part (8 MB); the i7-12700K has 25 MB. Substituting the real value gives `nc = 12800`, so `njb = 1` for any `n ≤ 12800` — the jc dimension collapses even harder than today, and `packB` becomes 26.2 MB against a 25 MB cache. Measured consequences of naively correcting `hw_traits` are in [#110](https://github.com/stillwater-sc/mtl5/issues/110).

## For context

[...]

Issue #222, opened 2026-07-19
perf(simd): runtime cache/ISA detection + scalable-vector (SVE/RVV) SIMD path


Ravenwater:
> **Tracker.** This bundles three independent workstreams, now split so each can
> be reviewed and scheduled on its own. This issue stays open as the umbrella
> until all three land.
>
> | # | workstream | status |
> |---|---|---|
> | 1 | Runtime cache detection feeding the GEMM blocking model | PR #426 |
> | 2 | Scalable-vector (SVE/RVV) SIMD path | #427 — blocked on hardware |
> | 3 | Highway runtime ISA dispatch | #428 — build-model decision |
>
> Original text below, unchanged.

## Problem

The SIMD/GEMM stack is fast but not portable across hardware:

1. **Blocking parameters are hardcoded to a Haswell-class core**
   (`include/mtl/simd/blocking.hpp` — 32KB L1 / 256KB L2 / 8MB L3, FMA latency 4,
   2 units). There is no runtime cache detection, so mr/nr/kc/mc/nc are
   suboptimal on other microarchitectures.
2. **No SIMD on scalable-vector ISAs.** ARM SVE and RISC-V V fall back to the
   scalar path because Highway's scalable vectors have no compile-time lane count
   (`include/mtl/simd/batch.hpp` demotes `HWY_HAVE_SCALABLE` to `size==1`).

## Goal

Make the existing kernels perform well beyond Haswell-class x86.

## Scope

- **Runtime cache/topology detection** (cache sizes, line size, core counts) that
  feeds the GEMM blocking model, replacing the hardcoded `hw_traits`. Keep the
  analytical BLIS model; just parameterize it at runtime (or per-build).
- **Scalable-vector SIMD path** for SVE / RVV — either a runtime-length `batch`
  variant or Highway's scalable API, so ARM SVE and RISC-V V get real
  vectorization instead of the scalar fallback.
- Consider Highway **runtime dispatch** (multiple ISA targets in one binary)
  instead of the current static single-ISA build, so a portable binary picks the
  best available ISA.

## Acceptance criteria

- GEMM/GEMV/L1 blocking adapts to the detected cache hierarchy (validated on a
  non-Haswell target).
- SVE and/or RVV builds vectorize (measurable speedup over the scalar fallback).
- No regression on the current x86/NEON paths.

**Effort:** moderate. Independent of the other roadmap items. Part of the
scale-out roadmap.

PR #382, opened 2026-08-04
perf(simd): size the GEMM register tile against the register file, not the latency floor

### LINKED ISSUE #351: perf(simd): single-core GEMM sustains ~72-78% of FMA peak — identify the microkernel limiter

Ravenwater:
## Summary

**+16% single-core GEMM** (fp64, i7-12700K, one pinned P-core) from a parameter-derivation change. No kernel was rewritten.

The tile came from Eq. 1 of the BLIS analytical model —

```
area = Nvec * Lvfma * Nvfma          // 4 * 4 * 2 = 32 elements
```

— used as the **target**. It is a **lower bound**: the minimum accumulators needed to cover dependent-FMA latency at the issue width. The derivation had **no register-file term at all**, so nothing modelled the ceiling. On AVX2 that gives 4×8 for double: **8 accumulator vectors out of 16 ymm, half the file idle**.

## The sweep

`mr`/`nr` are reachable by varying `fma_latency`/`fma_units`, the only handle the current model exposes. Single core, pinned, fp64, median of 3:

| tile | acc vectors | N=1024 | N=2048 |
|---|---:|---:|---:|
| **4×8** | 8 | 57.90 | 56.90 | ← shipped default |
| 5×8 | 10 | 63.41 | 64.16 |
| **6×8** | **12** | **67.57** | **66.96** | ← **+17%** |
| 8×8 | 16 | 57.68 | 57.03 | ← fills the file, spills |
| 6×12 | 18 | 60.24 | 59.38 |
| 8×12 | 24 | 55.31 | 55.72 |
| 11×12 | 33 | 53.33 | 54.31 |

**Both directions cost.** Too few accumulators cannot hide FMA latency; too many leave nothing for the operands and the microkernel spills every k step. The peak is at **~¾ of the register file** — which is also where **BLIS's hand-written Haswell `dgemm` kernel sits: 6×8**, the same shape this sweep found independently.

That is a sharper diagnosis than "derived at the minimum": the model was missing a constraint, not merely tuned low.

## Measured after implementing

| | before | after | |
|---|---:|---:|---|
| T=1 N=512 | 59.36 | **69.23** | +16.6% |
| T=1 N=1024 | 59.40 | **68.17** | +14.8% |
| T=1 N=2048 | 57.21 | **66.61** | +16.4% |
| T=8 N=2048 | 404.56 | **419.71** | +3.7% |

At N=2048 that moves single-core from **73% → 86% of FMA peak**, and from 75.5% → ~91% of OpenBLAS.

**The threaded gain is much smaller, and that is worth stating plainly:** at ~420 GFLOP/s the cores outrun what memory can feed, so parallel efficiency *drops* from 89% to 78% even as total throughput rises. Net win at every thread count, no regression — but this closes a single-core gap, not a scaling one.

## A bug I introduced and caught

`vec_registers` is **appended** to `hw_traits`, not inserted.

I inserted it third at first. `hw_traits` is a public aggregate that tests — and potentially downstream code — initialize **positionally**, so every later field silently shifted: a 32 KB `l1_bytes` became **8 bytes**, and `mr` came out as **12288**. The existing test caught it; user code would have been silently mis-configured with no diagnostic.

Appending means an initializer written before this field value-initializes it to `0`, which `derive_blocking` treats as *unknown* and falls back to the pre-#351 near-square latency-floor tile. **Old code keeps compiling and keeps its old behaviour** instead of getting garbage. The reasoning is a comment on the field so the ordering is not "tidied" later.

## Scope of the evidence

Measured **only on AVX2 with 16 ymm registers.** The ¾ budget extrapolates to 32-register ISAs (AVX-512, NEON, SVE) as 12×16 = 24 of 32, which is consistent with BLIS's AVX-512 kernel sitting at 24/32 — but that is *reasoning, not measurement*. ARM64 CI will confirm it still passes; it will not confirm it is optimal there.

#222 (runtime cache/ISA detection) is where a real `vec_registers` would come from rather than a compile-time default.

## Related

- **#347** — re-framed. Its premise ("#110 needs this before option 1") died with #110. `packB` is invariantly 8.00 MB across every tile in this sweep, so it did not confound these numbers; whether panel sizing contributes separately to the single-core gap is still open.
- **#381** — filed from this work: three MT sparse tests fail under `-march=native` because they compare bit-exactly across two different loop nests. Unrelated to this change; found while sweeping.

## Test results

| target | gcc build | gcc test | clang build | clang test |
|---|---|---|---|---|
| `simd_test_blocking` | OK | PASS (149 assertions / 8 cases) | OK | PASS |
| full unit suite | OK | **161/161** | OK | **161/161** |

New cases pin the register-file budget (12 of 16, with room for the B panel and A broadcast), assert the tile now clears the latency floor rather than sitting on it, and cover both the register-poor fallback and a 32-register file.

## Test plan
- [ ] Fast CI passes (gcc + clang, incl. ARM64)
- [ ] Promote to ready when satisfied

Resolves #351

### DIFF
--- a/include/mtl/simd/blocking.hpp
+++ b/include/mtl/simd/blocking.hpp
@@ -46,17 +46,26 @@ struct hw_traits {
     std::size_t l2_bytes;      // per-core L2
     std::size_t l3_bytes;      // shared L3 (per core group)
     std::size_t page_bytes;    // page size (TLB reasoning)
+    // APPENDED, not inserted (#351). hw_traits is a public aggregate and is
+    // initialized positionally in tests and downstream code; inserting a field
+    // mid-struct silently shifts every later one -- during this change that
+    // turned an l1_bytes of 32 KB into 8 bytes with no diagnostic. Appending
+    // means an initializer that predates this field value-initializes it to 0,
+    // which derive_blocking treats as "unknown" and falls back to the
+    // latency-floor tile, i.e. the pre-#351 behaviour.
+    std::size_t vec_registers; // Nreg: architectural vector registers (0 = unknown)
 };
 
 /// Generic modern-x86 (AVX2-class) default; override per architecture.
 /// Matches a Haswell-class core: 32 KB/8-way L1, 256 KB L2, 8 MB L3,
-/// FMA latency 4, 2 FMA units.
+/// FMA latency 4, 2 FMA units, 16 ymm registers.
 inline constexpr hw_traits default_hw_traits{
     /*fma_latency*/ 4, /*fma_units*/ 2,
     /*l1_bytes*/ 32u * 1024, /*l1_assoc*/ 8, /*line_bytes*/ 64,
     /*l2_bytes*/ 256u * 1024,
     /*l3_bytes*/ 8u * 1024 * 1024,
     /*page_bytes*/ 4096,
+    /*vec_registers*/ 16,
 };
 
 /// GEMM blocking parameters: mr x nr register microtile, kc/mc/nc cache blocks.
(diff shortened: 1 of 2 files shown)


Ravenwater:
@coderabbitai review

Issue #394, opened 2026-08-05
bug(itl): adjoint_solve is a stub in ilu_0/ssor/block_diagonal — wrong whenever the preconditioner is non-symmetric


Ravenwater:
Split out of #392, which reported it as a second finding. Independent of that issue's rho fix — confirmed by measuring after that fix landed.

## What is wrong

`bicg` and `qmr` are the only two solvers that call `adjoint_solve`; they need `M^-T` (or `M^-H`) applied to the shadow residual. Several preconditioners implement it by delegating to `solve`, which is valid **only if the preconditioner is self-adjoint**.

Surveying all of them:

| preconditioner | `adjoint_solve` | valid? |
|---|---|---|
| `identity` | `solve` | **yes** — I is self-adjoint |
| `diagonal` | `conj(inv_diag) * b`, a real implementation | **yes** |
| `ic_0` | `solve` — "L·Lᵀ is self-adjoint for real" | **yes** — IC is symmetric by construction |
| `ildl` | `solve` — "L·D·Lᵀ is symmetric" | **yes** — by construction |
| **`ssor`** | `solve` — "SSOR is symmetric" | **only if A is symmetric.** SSOR of a non-symmetric A is not symmetric |
| **`ilu_0`** | `solve` — "approximate" | **no.** M = L·U, so M^H = U^H·L^H — the triangular factors must be applied in the opposite order, transposed |
| **`block_diagonal`** | `solve` — "approximate" | **no**, unless every block is symmetric |

So `ic_0` and `ildl` are correct for a good reason (their factorizations are symmetric by construction), and the comments on `ssor`/`ilu_0`/`block_diagonal` are the ones that do not hold in general.

## Measured

`bicg` on current `main` (with #392's rho fix), 144×144, tolerance 1e-12, cap 3000:

| A | identity | diagonal | ssor | ilu_0 | ic_0 |
|---|---|---|---|---|---|
| symmetric | 35 ✓ | 33 ✓ | 17 ✓ | 12 ✓ | 12 ✓ |
| **non-symmetric** | 39 ✓ | 36 ✓ | **3000, rel 2.68e+00 ✗** | **3000, rel 8.53e+00 ✗** | — |

The dividing line is the **preconditioner's** symmetry, not the matrix's. `diagonal` works on a non-symmetric A because D is symmetric whatever A is; `ssor`/`ilu_0` of a non-symmetric A are themselves non-symmetric and fail.

[...]

PR #404, opened 2026-08-05
fix(itl): start ssor sweeps from x=0 so it applies the SSOR operator

### LINKED ISSUE #398: bug(itl): ssor starts its sweeps from x=b instead of x=0, so it is not the SSOR operator
**Rewritten.** The original framing of this issue was wrong: it claimed `ssor` needed reimplementing as "the classical factored operator", with a blast radius across every solver and a required before/after sweep. It is **one line**, and convergence *improves*.

Retained below is what the measurements actually show.

## The bug

`pc::ssor::solve` initialises its working vector to `b`:

```cpp
// Copy b into x as working vector
for (size_type i = 0; i < n_; ++i)
    x(i) = b(i);
```

then runs a forward and a backward SOR sweep. It should initialise to **zero**.

## Why zero

A stationary iteration and a preconditioner are the same matrix viewed two ways. SOR is defined by the splitting `A = M − N` with

```
M_SOR = (1/w)(D + wL),      iteration:  x <- x + M^-1 (b − A x)
```

and that `M` **is** the preconditioner. The consequence:

> Applying `M^-1` to a vector means running one sweep **starting from x = 0**.

From any other start you get `M^-1 b + G x0`, where `G = I − M^-1 A` is the iteration matrix. As a *solver* the initial guess is arbitrary — which is why the textbook presentations of SOR never mention it. As a *preconditioner* it must be zero, because the caller is asking for `M^-1 r`, not for an improved guess.

## The two sweeps ARE the factored operator

For SSOR the composite splitting matrix is

```
M_SSOR = w/(2−w) * (D/w + L) D^-1 (D/w + U)
```

This is not a different algorithm from the forward+backward sweeps — it is *what they compute*, when started from zero. At `w = 1`:

```
forward from 0:  (D+L) x1 = b            ->  x1 = (D+L)^-1 b
backward:        (D+U) x2 = b − L x1     ->  x2 = (D+U)^-1 (I − L(D+L)^-1) b
     and         I − L(D+L)^-1 = (D+L−L)(D+L)^-1 = D (D+L)^-1
     so          x2 = (D+U)^-1 D (D+L)^-1 b = [(D+L) D^-1 (D+U)]^-1 b
```

The original issue treated the factored form as a rewrite to be argued for. It is the same operator; only the starting vector differs.

## Why this destroys self-adjointness

If `A = A^T` then `U = L^T`, so `M = c (D/w + L) D^-1 (D/w + L)^T` — the form `X D^-1 X^T`, **symmetric by construction**. That is the whole reason the symmetric variant exists: Sheldon introduced SSOR precisely so the splitting matrix would be symmetric and usable with CG.

Starting from `x0 = b` makes the operator `M^-1 + G_B G_F`. Even with `A` and `M` both symmetric, `M^-1 A` is not — a product of symmetric matrices is symmetric only if they commute — so the composite is not symmetric. That extra term is the entire defect.

## Measured

Dense reference implementation, isolating only the initial vector:

| A | w | from `x0 = b` | from `x0 = 0` |
|---|---|---|---|
| symmetric | 1.0 | 7.26e-03 | **5.06e-16** |
| symmetric | 1.3 | 9.30e-02 | **5.00e-16** |
| non-symmetric | 1.0 | 5.44e-01 | 4.05e-01 |

and on MTL5's own `pc::ssor` with the initialisation changed:

| | before | after |
|---|---|---|
| adjoint violation, symmetric A | 1.1e-01 | **3.4e-16** |
| `cg + ssor` | 17 iterations | **13** |
| `gmres + ssor` | 13 iterations | **12** |

Self-adjoint to machine precision, and **convergence improves**. Non-symmetric A stays non-self-adjoint (2.6e-01), which is correct — `M` is not symmetric when `U != L^T`, and that is a property of the method, not a defect.

## What the original issue got wrong

- "It runs two SOR sweeps from `x = b` ... **not** the classical factored preconditioner." The first half is right, the second does not follow. The sweeps *are* the factored operator; the starting vector is the bug.
- "Reimplementing ... would change the iteration counts of every solver that uses it." Counts do change — `cg` 17 -> 13 — but in the right direction, and via a one-line fix rather than a rewrite.
- "This needs a before/after sweep, a decision on `omega` scaling, and confirmation that the multigrid smoothers still behave." The `omega` question is moot: `M_SSOR` carries a scalar factor `w/(2−w)` that different sources include or omit, and a scalar multiple changes neither symmetry nor the preconditioned system in any way that matters. Smoother behaviour is still worth checking, because `solve` also discards the incoming `x`.

## Note on the smoother use

`solve(x, b)` overwrites the caller's `x` before sweeping, so it is not a smoother step either — a smoother would start from the current iterate. With the fix it is unambiguously a preconditioner: `x <- M^-1 b`. Anything wanting the smoother semantics needs a separate entry point that respects the incoming `x`.

## References

- Young (1950), Harvard PhD thesis; Young (1954), *Trans. AMS* **76**, 92–111 — SOR. Frankel (1950), *MTAC* **4**, 65–75 — independent derivation.
- **Sheldon (1955)**, "On the numerical solution of elliptic difference equations", *MTAC* **9**, 101–112 — SSOR, introduced to symmetrise the splitting.
- Young (1971), *Iterative Solution of Large Linear Systems*.
- Barrett et al. (1994), *Templates*, §3.3 — gives `M_SSOR` explicitly.
- Saad (2003), *Iterative Methods for Sparse Linear Systems*, §4.1 and §10.2 — the preconditioner view.

Refs: #394 (where the non-self-adjointness was measured and the docs corrected), and `tests/unit/itl/test_preconditioner_adjoint.cpp`, the adjoint-identity test that will validate the fix.

Ravenwater:
## Summary

`pc::ssor` initialised its working vector to `b` before the forward and backward SOR sweeps. It must start from **zero**. One line in each of the two implementations.

A stationary iteration and a preconditioner are the same splitting matrix `M` viewed two ways: SOR is defined by `A = M − N`, and applying `M⁻¹` means running one sweep **from zero**. From any other start the result is `M⁻¹b + G·x₀` with `G = I − M⁻¹A` — fine for a solver, where the initial guess is arbitrary, wrong for a preconditioner, where the caller asked for `M⁻¹b`.

Run from zero the two sweeps are *exactly* the classical operator

```
M = w/(2-w) (D/w + L) D^-1 (D/w + U)
```

of Sheldon (1955), scalar factor included. Sweeps and triangular solves are the same operator here, not two competing formulations — which is why this is a one-line fix and not the reimplementation the issue originally described. Derivation is in the header, with `F = D/w + L`, `B = D/w + U`:

```
forward from 0:  F x1 = b
backward:        B x2 = b + (B - A) x1
so               x2 = B^-1 ((B - A) + F) F^-1 b,   (B - A) + F = D (2-w)/w
giving           x2 = [ w/(2-w) F D^-1 B ]^-1 b
```

For Hermitian `A`, `U = Lᴴ` and `D` is real, so `M = c·X·D⁻¹·Xᴴ` — self-adjoint by construction. That is precisely why Sheldon introduced the symmetric variant, and it makes `adjoint_solve`'s delegation to `solve` correct rather than a stub.

## Measured

144×144 Laplacian, adjoint identity `<M^-1 b, c> == <b, M^-H c>`:

| A | omega | before | after |
|---|---|---|---|
| symmetric | 1.0 | 1.119e-01 | **3.432e-16** |
| symmetric | 1.4 | 3.176e-01 | **3.827e-16** |
| non-symmetric | 1.0 | 4.105e-01 | 2.608e-01 |
| non-symmetric | 1.4 | 4.003e-01 | 3.744e-01 |

Iteration counts, same matrix, `rtol = 1e-12`:

| solver | symmetric A: before -> after | non-symmetric A: before -> after |
|---|---|---|
| `cg + ssor` | 16 -> **13** | 45 -> **does not converge** |
| `gmres + ssor` | 13 -> **12** | 13 -> 13 |
| `bicg + ssor` | 16 -> **13** | 3000 (no) -> 3000 (no) |
| `qmr + ssor` | 16 -> **13** | 121 (breakdown) -> 3000 (no) |
| `cg + diagonal` (control) | 34 -> 34 | 3000 (no) -> 3000 (no) |

Every solver improves on symmetric `A`, which is ssor's domain.

**One configuration gets worse and is worth stating plainly**: `cg + ssor` on a *non-symmetric* `A` converged in 45 iterations before and no longer converges. CG requires an SPD operator, so this was never a supported combination — its former convergence rested on the extra `G_B·G_F` term rather than on the preconditioner. No configuration inside the method's domain regresses.

SSOR of a non-Hermitian `A` remains non-self-adjoint (2.6e-01) because `U != Lᴴ`. That is a property of the method, not a residual defect, and is now asserted as such so it is not later "fixed".

## Changes

- `include/mtl/itl/pc/ssor.hpp` — zero-init in both the generic and `compressed2D` forms. Header note rewritten with the derivation, the Hermitian argument, and references; the two `adjoint_solve` docs replaced (they previously described the operator as "not the classical factored form", which is the claim this PR retires). Also records that `solve()` overwrites `x`, so this is a preconditioner and **not** a smoother — the smoother is a separate class, `itl::smoother::symmetric_sor`, which is what the multigrid code uses. That is why the blast radius here is four test files and no solver behaviour outside ssor.
- `tests/unit/itl/test_advanced_pc.cpp` — new: `pc::ssor` output vs a dense LU solve of `M`, at `w = 1.0` and `1.3`. `w = 1` collapses the scalar factor to 1 and would not catch a wrong one, so the second omega carries that part of the claim.
- `tests/unit/itl/test_preconditioner_adjoint.cpp` — new: adjoint identity for ssor at two omega, plus an explicit assertion that a non-symmetric `A` is **not** self-adjoint.
- `CHANGELOG.md`

## Negative control

Reverting **only** the header, keeping both new tests:

```
test_advanced_pc.cpp:233: FAILED: REQUIRE( rel < 1e-12 )
  omega = 1, ||x - M^-1 b|| / ||M^-1 b|| = 0.24471
test_preconditioner_adjoint.cpp:144: FAILED: ssor, symmetric A, violation = 0.111907
test_preconditioner_adjoint.cpp:153: FAILED: ssor, symmetric A, omega = 1.4, violation = 0.317625
```

The non-symmetric assertion passes in both worlds, as it should.

## Test Results

| | build | ctest |
|---|---|---|
| gcc 13, Debug | OK | 163/163 |
| clang, Debug | OK | 163/163 |

## References

Young (1954) *Trans. AMS* **76**, 92–111 and Frankel (1950) *MTAC* **4**, 65–75 (SOR); **Sheldon (1955)** *MTAC* **9**, 101–112 (SSOR); Young (1971) *Iterative Solution of Large Linear Systems*; Barrett et al. (1994) *Templates* §3.3; Saad (2003) §4.1, §10.2.

Resolves #398

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected SSOR sweeps to start from zero, matching the classical preconditioner behavior.
  * Improved SSOR handling for Hermitian and non-Hermitian matrices.
  * Updated adjoint behavior for real relaxation factors.

* **Tests**
  * Added regression coverage for multiple relaxation factors.
  * Added validation for symmetric and non-symmetric matrix behavior.


Ravenwater:
Both review findings addressed in 8b1454b.

**Namespace cross-reference** — taken. One correction to the rationale though: `itl::smoother::symmetric_sor` *does* resolve from inside `namespace mtl::itl::pc`, since unqualified lookup of `itl` walks out to the enclosing `mtl` and finds `mtl::itl` there. So it was not wrong as C++. It is still worth qualifying, because a comment gets read out of context, and that is why I applied it. The identical short forms in `tests/unit/itl/test_smoothers.cpp` are real code under `using namespace mtl;` and are left alone.

**Named SECTION** — taken, and extended to all three cases in that test case rather than only the reverse-condition one, so the three read consistently. Also recorded at the site that the third asserts a *lower* bound on an error, and that if it ever fires the right response is to delete the section rather than loosen the bound.

Worth noting explicitly since it cuts against the file's other test case: SECTIONs are safe here only because there is no enclosing loop. `preconditioner adjoint_solve really is the adjoint (#394)` right above deliberately uses bare blocks inside a `for (bool sym : ...)` loop, for the reason stated in its own comment — Catch2 re-enters the whole test case per section, which interacts badly with a loop and obscures which case failed. I left a note in the source so the two are not later "made consistent" in the wrong direction.

163/163 on gcc and clang.

PR #406, opened 2026-08-05
refactor(itl): make pc::ssor the adapter over smoother::symmetric_sor

### LINKED ISSUE #405: refactor(itl): make pc::ssor the adapter over smoother::symmetric_sor it is documented to be
## What

`pc::ssor` is documented as a thin adapter over `smoother::symmetric_sor`. It is not one — it is an independent reimplementation of the same two sweeps. Make the code match the design.

`docs/design/smoothers-architecture.md` states the rule directly:

> **Smoothers expose `operator()`, not `solve()`.** They are not themselves `Preconditioner`-concept types; the `pc::ssor` wrapper adapts the symmetric sweep to the `solve`/`adjoint_solve` face when a Krylov method needs one.

and again under *Where they are used*:

> **Krylov preconditioning** uses the symmetric variants (via `pc::ssor`) as an SPD-preserving `M` for CG and friends.

But `include/mtl/itl/pc/ssor.hpp` never mentions `smoother::symmetric_sor`. It carries its own forward and backward sweep loops and its own diagonal cache, duplicated across both the generic and the `compressed2D` specialization — roughly 120 lines restating `include/mtl/itl/smoother/sor.hpp`.

## The split is right; only the implementation is duplicated

The two types have genuinely different contracts, and that distinction should stay:

| | `smoother::symmetric_sor` | `pc::ssor` |
|---|---|---|
| Face | `operator()(x, b)` | `solve` / `adjoint_solve` |
| Semantics | `x <- x + M^-1 (b - A x)` — *relaxes* an iterate | `x <- M^-1 b` — *applies* the operator |
| Incoming `x` | **load-bearing** | **overwritten** |
| Called | `nu_pre`/`nu_post` times per grid level | once per Krylov iteration |
| Concept | not a `Preconditioner` | `Preconditioner` |

That is exactly the distinction #398 turned on. What makes the refactor safe is that after #404 the two are provably the same operator:

    pc::ssor::solve(x, b)  ==  x = 0; symmetric_sor(x, b)

Measured on a 144x144 Laplacian, relative difference:

| omega | `||pc::ssor - smoother_from_0|| / ||.||` |
|---|---|
| 1.0 | 7.960e-17 |
| 1.3 | 1.726e-16 |
| 1.7 | 3.059e-16 |

The sweep bodies are arithmetically identical; the only difference is `dia_inv_(i) * (...)` in the smoother against `(...) / dia_(i)` in the preconditioner, a reciprocal-multiply worth a last-ulp difference.

## Why it is worth doing

**Mixed precision is unavailable exactly where the library cares about it.** `smoother::sor<Matrix, Accumulator>` threads a wider or exact (quire) accumulator through both sweeps, with the omega blend deliberately left outside it. `pc::ssor` has no `Accumulator` parameter at all. So today you can *smooth* in mixed precision but you cannot *precondition* in it — and nothing in the API says so. The adapter inherits the parameter for free.

**The duplication is what made #398 possible.** `smoother::symmetric_sor` respects the incoming `x`, which is correct for a smoother, and was never wrong. `pc::ssor` had to restate the "start from a known vector" step itself, and started it from `b`. Had it been the three-line adapter — zero `x`, apply the smoother once — the bug could not have been written. Same pattern as #392/#393/#394/#403: the defect lived in a second copy of something already correct.

**It deletes ~120 lines** and removes a second place for the two to drift apart.

## Sketch

```cpp
template <typename Matrix, typename Accumulator = void>
class ssor {
    using value_type = typename Matrix::value_type;
public:
    explicit ssor(const Matrix& A, value_type omega = value_type(1))
        : n_(A.num_rows()), sm_(A, omega) {}

    template <typename VecX, typename VecB>
    void solve(VecX& x, const VecB& b) const {
        for (std::size_t i = 0; i < n_; ++i) x(i) = math::zero<value_type>();
        sm_(x, b);          // one application FROM ZERO is exactly M^-1 b
    }

    template <typename VecX, typename VecB>
    void adjoint_solve(VecX& x, const VecB& b) const { solve(x, b); }
private:
    std::size_t n_;
    mutable smoother::symmetric_sor<Matrix, Accumulator> sm_;
};
```

Both the generic and `compressed2D` specializations collapse into this one template, since `smoother::sor` already specializes on the matrix type.

## Wrinkles

- `symmetric_sor::operator()` and `sor::sweep` are **non-const**, while `pc::ssor::solve` is const. `sweep` mutates only `x`, never a member, so the clean fix is to make it const rather than to add `mutable` as sketched above. Worth checking `jacobi` and `gauss_seidel` for the same while there.
- The header comment on `pc/ssor.hpp` — the derivation of `M = w/(2-w)(D/w + L)D^-1(D/w + U)`, the Hermitian argument and the references — should survive the refactor. It is the reason the zero-init is not a detail.
- Keep the existing tests as-is: `test_advanced_pc.cpp` checks `pc::ssor` against a dense LU of `M`, and `test_preconditioner_adjoint.cpp` checks the adjoint identity. Both are behavioural and should pass unchanged; if they do not, the refactor is not equivalence-preserving.
- Adding `Accumulator` is an API addition, not a break — it defaults to `void`, matching today's arithmetic.

Follow-up to #398 / #404. Not urgent; the operator is correct today, this is about there being one implementation of it rather than two.

Ravenwater:
## Summary

`pc::ssor` was documented as a thin wrapper over `smoother::symmetric_sor` but was an independent reimplementation of the same two sweeps — its own loops and diagonal cache, duplicated across a generic and a `compressed2D` specialization. `docs/design/smoothers-architecture.md` already described the intended shape:

> **Smoothers expose `operator()`, not `solve()`.** They are not themselves `Preconditioner`-concept types; the `pc::ssor` wrapper adapts the symmetric sweep to the `solve`/`adjoint_solve` face when a Krylov method needs one.

This makes the code match. #398 established that one symmetric-SOR application from a zeroed vector *is* `M^-1 b`, so the preconditioner collapses to three lines over the smoother.

**−129 lines** in `pc/ssor.hpp` (217 → 115, most of the remainder being the derivation and references). One class definition replaces two: the matrix-type specialization lives in `smoother::sor`, so the O(nnz) sparse path is inherited rather than restated.

## Two gains beyond the line count

**Mixed-precision preconditioning becomes possible.** `pc::ssor` now takes the `Accumulator` parameter and forwards it. Previously you could *smooth* in mixed precision but not *precondition* in it, and nothing in the API said so — the parameter simply did not exist on the preconditioner. It defaults to `void`, so this widens the old single-parameter interface rather than breaking it; `pc::ssor<mat::compressed2D<double>>` still compiles unchanged.

**The #398 bug becomes unrepresentable.** That bug (`x = b` instead of `x = 0`) lived in the duplicate. `smoother::symmetric_sor` respects the incoming `x`, which is correct for a smoother and was never wrong. Now zeroing `x` is the only thing separating the preconditioner from the smoother, and it is one visible line with the reason next to it.

## Const-correctness

The smoothers' sweeps are now `const`. They only ever mutated the caller's `x` — `A_`, `omega_` and `dia_inv_` are read-only during a sweep — so this is what they always were. It is what lets `pc::ssor` hold a smoother as a member and still expose a `const solve()` without reaching for `mutable`. Applied to `jacobi` and `gauss_seidel` too, for consistency. Widening a method to `const` is not a breaking change.

## Equivalence

Carried by tests already in place rather than by inspection — this is the payoff for having landed them in #404 first:

- `test_advanced_pc.cpp` checks `pc::ssor` against a dense LU of `M = w/(2-w)(D/w+L)D^-1(D/w+U)` at two omega.
- `test_preconditioner_adjoint.cpp` checks the adjoint identity at two omega, both directions.

Both pass **unchanged**. Iteration counts identical:

| | before | after |
|---|---|---|
| `cg + ssor` | 13 | 13 |
| `gmres + ssor` | 12 | 12 |
| `bicg + ssor` | 13 | 13 |
| `qmr + ssor` | 13 | 13 |

Adjoint violation moves 3.432e-16 → 3.545e-16 — the expected last-ulp shift from the smoother's `dia_inv_ *` against the old `/ dia_`. The one figure quoted in the header comment was updated to match.

## Changes

- `include/mtl/itl/pc/ssor.hpp` — the adapter. Header documentation from #398 (derivation, Hermitian argument, Sheldon/Young/Saad references) preserved; a note added on why there is one implementation rather than two. Includes narrowed to what the header now uses.
- `include/mtl/itl/smoother/sor.hpp` — sweeps `const`, with a note on why.
- `include/mtl/itl/smoother/gauss_seidel.hpp`, `jacobi.hpp` — same.
- `tests/unit/itl/test_advanced_pc.cpp` — new test that the accumulator is actually *invoked* through the preconditioner (not merely accepted as a parameter), that the result still satisfies `M x = b`, and that the default `void` accumulator does not touch it.
- `CHANGELOG.md`

## Test Results

| | build | ctest |
|---|---|---|
| gcc 13, Debug | OK | 163/163 |
| clang, Debug | OK | 163/163 |

Resolves #405

## Summary by CodeRabbit

- **New Features**
  - SSOR now supports configurable accumulator precision for improved numerical flexibility.
  - SSOR behavior remains consistent while using shared smoothing operations for forward and backward sweeps.
  - Smoothing operations can now be invoked through const-qualified instances.

- **Documentation**
  - Updated the unreleased changelog and technical documentation to reflect SSOR and smoothing behavior.

- **Tests**
  - Added coverage verifying custom accumulator support and computed preconditioner results.


Ravenwater:
Both nitpicks taken in aeba95d.

**Exact call count** — this one is better than a nitpick and worth saying so. `solve()` now *delegates* to `symmetric_sor` rather than running two visible loops, so silently losing one of the two sweeps is precisely the regression this refactor makes possible, and `> 0` sails straight past it. Verified rather than assumed: dropping the backward sweep gives

```
test_advanced_pc.cpp:222: FAILED:
  REQUIRE( g_ssor_addproduct_calls == static_cast<int>(4 * (n - 1)) )
  add_product calls = 30
```

against the expected 60, and the assertion catches it. The proposed `4*(n-1)` is right — `add_product` fires once per *stored* off-diagonal entry (the diagonal is skipped by index, not by value), this tridiagonal `A` has `2*(n-1)` of them, and `symmetric_sor` runs the pair. I kept the derivation in the source so the constant is checkable rather than magic, and left an `INFO` so a future failure reports the actual count.

**1x1 case** — taken as a separate `TEST_CASE`. Your reasoning is the useful part: it is the one path where the accumulator yields a result without ever accumulating, so `sigma` comes from `AT::value` on a freshly cleared accumulator and a `clear()`/`value()` disagreement about what an empty sum means would be invisible at every other size. Also asserts the accumulator and default-`void` paths agree on the answer (`M^-1 b = 1/4` exactly, since `M = (D+L)D^-1(D+U)` collapses to `D`), so neither is merely self-consistent.

163/163 on gcc and clang.

PR #407, opened 2026-08-05
feat(mat): sparse * sparse product (SpGEMM), Gustavson's algorithm

### LINKED ISSUE #402: feat(mat): sparse * sparse returns dense2D — Galerkin coarse operators R*A*P materialise dense
Split out of #401, which flagged it as an aside while fixing the multigrid include. Confirmed, and it is a capability gap rather than a defect.

## What happens

```cpp
using SMat = mtl::mat::compressed2D<double>;
SMat A(4,4), B(4,4);
auto C = A * B;   // -> mtl::mat::dense2D<double>
```

Verified at `dce9bb0`: `compressed2D * compressed2D` yields **`dense2D`**.

## Why it matters where it was found

`itl::mg::multigrid`'s constructor asks the caller for a hierarchy of level matrices, and the standard way to build the coarse levels is the **Galerkin triple product**

```
A_coarse = R * A * P
```

with all three sparse. Today that materialises a dense intermediate at the fine-grid size. For a 1-D problem with n = 63 that is irrelevant; for the 2-D and 3-D problems multigrid exists to solve, an `n x n` dense intermediate is the whole point of not using a dense solver.

So the one product a multigrid user most needs to stay sparse is the one that does not.

## Not a correctness bug

The result is numerically correct — `dense2D` holds the right values. Nothing is wrong; the sparse-times-sparse-to-sparse product simply appears not to be implemented, and `operator*` falls back to a dense result. Filing it as a feature request rather than a bug for that reason.

## What an implementation needs

The classical approach is Gustavson's algorithm: for each row `i` of the left operand, scatter the linear combination of the right operand's rows into a sparse accumulator, then gather the nonzeros into the result row. MTL5 already has the accumulator — `sparse/util/scatter.hpp`, described in the design doc as "the workhorse of sparse arithmetic" — so the missing piece is the row-by-row product on top of it, not new infrastructure.

Two decisions worth making deliberately:

- **Symbolic pass or single pass?** A separate symbolic phase gives exact result sizing and reusable structure when the same pattern is multiplied repeatedly, which is exactly the multigrid setup case; a single pass with a growing accumulator is simpler and fine for one-shot use.
- **Should `operator*` change return type?** Making `sparse * sparse -> sparse` changes the type of an existing expression. Anything currently relying on the dense result would keep compiling only if it accepts both, so this wants a check of the call sites, or a separately named entry point (e.g. `spgemm`) with `operator*` following later.

## Refs

#401 (where this was observed, now fixed for the include problem it was an aside to). `sparse/util/scatter.hpp` for the accumulator. `docs/sparse-direct-solvers-design.md` lists the scatter as the intended basis for sparse arithmetic.

Ravenwater:
## Summary

`compressed2D * compressed2D` fell through to the generic dense `operator*`, which materialised an `n x n` dense intermediate **at the fine-grid size** and reached every element through `compressed2D::operator()(r, c)` — itself a binary search. The Galerkin triple product `A_coarse = R * A * P`, the one product a multigrid user most needs to stay sparse, was the one that did not.

`mtl::spgemm(A, B)` implements Gustavson (1978) row-by-row over the existing sparse accumulator (`sparse/util/scatter.hpp`, CSparse's `cs_scatter`), so work is proportional to the multiply-adds actually performed rather than `n^3`. The infrastructure was already there — this is the row-by-row product on top of it, as the issue anticipated.

## The two decisions the issue asked for

**`operator*` changes return type** for two `compressed2D`. I verified before making the change that no call site in the repository multiplies two sparse matrices, so nothing relied on the dense result — the `A * B` uses in tests are all `dense2D`. Partial ordering picks the overload, the same mechanism by which the existing `compressed2D * dense_vector` overload wins over `operator*(M, V)`. A `static_assert` in the tests pins the return type so a silent regression to the dense path is a compile error rather than a slow test.

**Single pass, no symbolic phase.** A symbolic pass buys exact result sizing and a reusable pattern when the *same* structure is multiplied repeatedly. Multigrid setup builds each level once, and `std::vector` growth is amortised O(1), so building it now would be speculative. The header notes where it would slot in.

Also decided: entries that cancel to exactly zero are **kept**. A Galerkin hierarchy wants a pattern stable across a nonlinear solve's re-assemblies, not one that shifts with the numbers.

## Sorting, and why the tests look the way they do

`compressed2D::operator()(r, c)` locates entries with `std::lower_bound`, so ascending column indices per row are a **correctness invariant**, not a storage convention. The accumulator returns indices in scatter (first-touch) order, so each gathered row is sorted before it is written.

Worth recording how invisible skipping that would be. Measured with the sort removed, 12x12 at 30% density:

| checked how | result |
|---|---|
| `nnz` | **correct** |
| row sums from the raw CSR arrays | **correct to 4.4e-16** |
| sparse matvec `A*x` | **exactly correct** |
| element access `C(i,j)` | **wrong by 3.2 absolute** |

The matvec is exact because it sums `data[k]*x(indices[k])` in storage order, which does not care about order. So every natural test for a matrix product — residual of `A*x`, nnz, row sums — passes on a corrupt result. `test_spgemm.cpp` therefore asserts the index structure directly rather than trusting a residual.

Control confirms it: removing the sort fails 3 of 6 cases with `row 0: indices[0] = 1, indices[1] = 0`.

## Changes

- `include/mtl/operation/spgemm.hpp` — new. Gustavson row-by-row, gather buffer reused across rows.
- `include/mtl/mat/operators.hpp` — `operator*(compressed2D, compressed2D)` delegating to `spgemm`, plus the include.
- `tests/unit/operation/test_spgemm.cpp` — new. Rectangular product with an empty row (a square-only test cannot catch a result sized from the wrong operand); random matrices at two densities either side of heavy row overlap; `operator*` return type; the Galerkin triple product; `1x1` / structurally-empty / zero-sized edges; exact cancellation; mixed `float`/`double` widening.
- `CHANGELOG.md`

One note on the Galerkin test: I derived the coarse-operator constant rather than pattern-matching it, having first got it wrong — with `R = P^T/2` the product is `(1/2)P^T A P = tridiag(-0.25, 0.5, -0.25)`, and I initially asserted the un-halved `P^T A P` values against the halved `R`. The derivation is in the test so the constant is checkable.

## Test Results

| | build | ctest |
|---|---|---|
| gcc 13, Debug | OK | 164/164 |
| clang, Debug | OK | 164/164 |

## Test plan
- [ ] Fast CI passes
- [ ] Promote to ready when satisfied

Resolves #402

## Summary by CodeRabbit

* **New Features**
  * Added sparse matrix–matrix multiplication support.
  * Sparse multiplication now returns a sparse matrix, improving efficiency for sparse inputs.
  * Added support for rectangular matrices, mixed value types, cancellation cases, and Galerkin triple products.
  * Results maintain correct column ordering and preserve structural zero entries.

* **Bug Fixes**
  * Improved correctness of sparse multiplication results and element access.


Ravenwater:
Addressed in 8961072 — but the finding as stated is not correct, and the distinction matters for the next reader.

**The claim.** That when `nnz == 0` the raw-CSR constructor "can perform undefined null-pointer arithmetic" via `indices + nnz_count`.

**Why it does not hold.** Since C++17 ([expr.add]/4, resolving CWG 232), adding `0` to a null pointer yields a null pointer. So `indices + 0` is well-defined, `[null, null)` is an empty range, and the constructor's copies are no-ops. MTL5 is C++20 throughout. Verified rather than argued — all four empty shapes under `-fsanitize=undefined,address`:

```
empty operands: 3x2 nnz=0 starts=4
zero-sized:     0x0 nnz=0 starts=1
inner dim 0:    2x3 nnz=0 starts=3
n == 0:         2x0 nnz=0 starts=3
```

No diagnostics. The code was correct as written.

**Taken anyway, on a different merit.** The rule is recent enough that it did *not* always hold — it was arguably UB in C++11 — and relying on it required a standard-lawyer comment to explain the code to anyone reading it. An early return for `nnz == 0` says what is meant instead: the `(rows, cols)` constructor builds exactly the all-zero `starts_` the function just computed. The footnote goes away, which is worth more than the line it costs.

**And it surfaced a real gap**, which is the useful part of the finding. Reviewing the empty paths turned up two shapes the tests missed, now covered:

- **inner dimension zero**, `(2x0)*(0x3)` → `2x3`. The contraction is over nothing, so the result is shaped from the *outer* dimensions — a result sized off the wrong operand would get this wrong and nothing else in the suite would notice.
- **zero output columns**, `(2x3)*(3x0)`. This is the case the accumulator is sized `n ? n : 1` for; a zero-length workspace would be an invalid allocation to index into even though nothing is ever scattered.

164/164 on gcc and clang.

Issue #408, opened 2026-08-05
perf(simd): #382's register tile regresses multi-core GEMM — 8-thread efficiency 90.5% -> 79.2%, and -7.4% absolute at N=1024


Ravenwater:
Found while re-measuring the i7-12700K benchmark page after #382 (`docs/benchmarks/i7-12700k.md`, 2026-08-05 pass).

## What

#382 sized the GEMM register tile against the register file rather than the latency floor. Single-threaded that is a clear win — **+22%**, and the epic #82 gate moves from 82.3% to **97.4%** of OpenBLAS. But the gain decays monotonically with thread count, and at N = 1024 with 8 threads it goes **negative**.

native-fast, GFLOP/s, before (2026-08-03) → after (2026-08-05):

| T | N = 1024 | N = 2048 |
|---:|---:|---:|
| 1 | 56.28 → 68.38 (**+21.5%**) | 55.46 → 67.81 (**+22.3%**) |
| 2 | 114.24 → 125.24 (+9.6%) | 111.62 → 133.86 (+19.9%) |
| 4 | 225.54 → 233.67 (+3.6%) | 216.08 → 253.56 (+17.3%) |
| 8 | 388.90 → 360.18 (**−7.4%**) | 401.75 → 429.40 (+6.9%) |

## Consequences on the page

- **Parallel efficiency at 8 threads: 90.5% → 79.2%.** native-fast went from the *best* scaler of the four backends to the *worst*: BLIS 7.37×, MKL 7.21×, OpenBLAS 7.11×, native-fast **6.33×**.
- **Relative throughput is no longer flat across thread counts.** It was 75.5% of OpenBLAS at T=1 and 76.4% at T=8 — the basis for the page's claim that "threading is no longer what separates native-fast from the tuned libraries." It is now **91.5% at T=1 and 81.5% at T=8**, so the differentiator has moved from the single-core kernel to the parallel path. That claim is retracted on the page.

Both retractions are marked in place rather than silently edited.

## Not a reason to revert

The serial win is large and the gate result is real. This is a tuning question — the tile that is optimal for one thread is evidently not optimal for eight — not a defect in #382's reasoning.

## Hypothesis, explicitly untested

[...]

Issue #401, opened 2026-08-05
bug(mg): multigrid.hpp is not self-contained — uses A*x without including mat/operators.hpp


Ravenwater:
`mtl/itl/mg/multigrid.hpp` does not compile on its own. `vcycle` and `wcycle` form the residual with

```cpp
auto Ax = levels_[l] * x;      // multigrid.hpp:84, and again at :122
```

but the header's include list stops at `dense_vector.hpp`, `compressed2D.hpp`, `norms.hpp`, `restriction.hpp` and `prolongation.hpp` — none of which pull in `mtl/mat/operators.hpp`, where `operator*(const compressed2D&, const V&)` lives (`operators.hpp:128`).

So a translation unit that includes `mg/multigrid.hpp` and nothing else fails:

```
error: no match for 'operator*' (operand types are
  'mtl::mat::compressed2D<double>' and 'mtl::vec::dense_vector<double>')
```

It builds only if the caller happens to have included `mat/operators.hpp` first, which makes it order-dependent.

Verified at `dce9bb0`.

## Fix

```diff
 #include <mtl/vec/dense_vector.hpp>
 #include <mtl/mat/compressed2D.hpp>
+#include <mtl/mat/operators.hpp>
 #include <mtl/operation/norms.hpp>
```

With that added the header compiles standalone, and a two-level V-cycle behaves as it should — on 1-D Poisson (n = 63, Gauss-Seidel smoother, nu_pre = nu_post = 2, Galerkin coarse operator R·A·P) the residual drops by a factor of **0.144 per cycle**, steady across cycles:

```
start   ||r|| = 7.937e+00
cycle 1 ||r|| = 1.287e+00   factor 0.162
cycle 2 ||r|| = 1.855e-01   factor 0.144
cycle 3 ||r|| = 2.680e-02   factor 0.144
...
```

So this is purely the missing include; the algorithm is fine.

## Unrelated observation

`operator*` on two `compressed2D` operands returns a **dense2D**, so forming a Galerkin coarse operator `R * A * P` materialises a dense intermediate at the fine-grid size. For a multigrid hierarchy that is the one product you most want to stay sparse. Not a defect — the sparse × sparse → sparse product may simply not be implemented yet — but it makes `R * A * P` impractical at scale, and it is the natural way to build the coarse levels that `multigrid`'s own constructor asks for.

Issue #393, opened 2026-08-05
bug(itl): qmr fails with any non-identity preconditioner — applies the whole M where the algorithm splits it


Ravenwater:
Split out of #392, which reported it as a second finding. Not fixed by that issue's one-line change — verified.

## Symptom

Identical in shape to the bicg bug: converges with `pc::identity`, stalls with anything else.

```
symmetric positive definite A (n=144)
  cg   + diagonal (control)  info=0 iters=  33 rel_err=2.05e-12  ok
  qmr  + identity            info=0 iters=  35 rel_err=2.43e-12  ok
  qmr  + diagonal            info=1 iters=3000 rel_err=2.48e-01  ** WRONG **
```

The preconditioner is not in question — `diagonal.solve` returns `1/A(0,0)` exactly, and `cg` with the same preconditioner on the same matrix converges in 33.

## Not the same defect as #392

#392 fixed `bicg`'s rho, and the fix is confirmed necessary and sufficient there for symmetric preconditioners. **Applying it does not help qmr** — that was checked as part of resolving #392, with `qmr + diagonal` still at 3000 iterations / 2.48e-01 after `bicg + diagonal` had gone to 33 / 2.05e-12.

## Suspected cause, not yet isolated

QMR (Freund–Nachtigal) uses a **split** preconditioner `M = M1 · M2`, applying `M1^-1` on one side and `M2^-1` on the other. `krylov/qmr.hpp:40-41` applies the whole `M` in both places:

```cpp
M.solve(y, v_tilde);          // expects M1^-1
M.adjoint_solve(z, w_tilde);  // expects M2^-T
```

With `identity` the distinction vanishes, which fits the observed pattern exactly. But I have **not** isolated it to that line, and the residual/scaling recurrences downstream would need checking too — a split-preconditioner error can also show up in the `rho`/`xi`/`gamma` updates. Filing what is measured rather than guessing at the patch.

## Why it survived

`tests/unit/itl/test_qmr.cpp` uses `pc::identity` in **both** of its test cases — the one configuration where the defect is invisible. The same was true of `test_bicg.cpp` before #392.

The correlation across the solver suite is exact:

| solver | preconditioners in its tests | status |
|---|---|---|
| `bicg` | identity only | was broken (#392) |
| **`qmr`** | **identity only** | **broken** |
| `bicgstab` | identity **and diagonal** | correct |

[...]

Issue #392, opened 2026-08-05
bug(itl): bicg computes rho from the wrong vectors — fails with any non-identity preconditioner


Ravenwater:
`mtl::itl::bicg` converges only when the preconditioner is `pc::identity`. With any real preconditioner it stalls and returns a solution that is wrong by O(1) — including a `pc::diagonal` on a symmetric positive definite matrix, where nothing about the preconditioner is in question.

Verified at `0791e5b`, found while building a solver/preconditioner dispatch for the Python bindings.

## Cause

`krylov/bicg.hpp:46` and `:90`:

```cpp
value_type rho = mtl::dot<Accumulator, value_type>(z_tilde, z);
```

Preconditioned BiCG requires the preconditioned residual against the **un**preconditioned shadow residual (Barrett et al., *Templates*, §2.3.5; Saad, *Iterative Methods*, Alg. 7.1):

```
rho_{i-1} = z_{i-1}^T r~_{i-1}
```

MTL5 uses `z~ · z`, which applies the preconditioner twice: `z~ · z = (M^-T r~) · (M^-1 r) = r~^T M^-1 M^-1 r`, where the algorithm needs `r~^T M^-1 r`. With `identity`, `z == r` and `z~ == r~`, so the two coincide — which is exactly why the identity case is the only one that works.

## Fix

One line, in both places:

```diff
-    value_type rho = mtl::dot<Accumulator, value_type>(z_tilde, z);
+    value_type rho = mtl::dot<Accumulator, value_type>(r_tilde, z);
```

Verified. On the reproducer below, `bicg + diagonal` goes from *3000 iterations, rel_err 2.25e-01* to **33 iterations, rel_err 2.05e-12** — matching `cg + diagonal`'s 33 on the same system, which is the right answer for an SPD matrix.

## Reproducer

```cpp
#include <cstdio>
#include <cmath>
#include <random>
#include <mtl/mat/compressed2D.hpp>
#include <mtl/mat/inserter.hpp>
#include <mtl/vec/dense_vector.hpp>
#include <mtl/operation/mult.hpp>
#include <mtl/itl/iteration/basic_iteration.hpp>
#include <mtl/itl/pc/identity.hpp>
#include <mtl/itl/pc/diagonal.hpp>
#include <mtl/itl/krylov/bicg.hpp>
#include <mtl/itl/krylov/qmr.hpp>
#include <mtl/itl/krylov/cg.hpp>

[...]

PR #413, opened 2026-08-06
docs: session log for 2026-08-05, and fill the CHANGELOG gaps

Ravenwater:
## Summary

Session log for 2026-08-05, plus six missing `Unreleased` CHANGELOG entries.

## Session log

`docs/sessions/2026-08-05-every-diagnosis-was-wrong.md`. The title is the honest through-line: **every up-front diagnosis offered today was wrong**, and a short measurement said so each time.

| Claimed | Actually |
|---|---|
| `pc::ssor` needs a reimplementation with wide blast radius (#398, as filed) | One line — and it *improves* convergence |
| #382 regressed scaling via cache pressure from the larger tile (#408, as filed) | `mc` was coupled to `mr` |
| MKL's small-N anomaly is a cold P-core at 800 MHz | Refuted: 60.8 GFLOP/s cold vs 61.0 warm |
| "164/164 passing" after a change | Build had **failed**; ctest passed a stale binary |
| A comparison showing +5% at T=1 | Cross-session drift; that path *cannot* change |

It records what went wrong as well as what shipped:

- **A silent-wrongness bug I introduced** in the #410 fix — `serial_nest` stepping `ic += MC` while blocks were sized `min(MC_eff, …)`, dropping rows of C. The entire existing GEMM suite passed it, because every case used an `m` fitting one ic-block.
- **A false green I reported** — "164/164" against a stale binary after a failed build; the signal was a `grep -c` returning 1 that I under-weighted.
- **Three watchers that looked armed and could not fire** — two CI monitors on `jq` (not installed here), two process waiters on the `pgrep -f` self-match trap that was *already in my notes from a previous session* and still hit twice.

## CHANGELOG gaps

Six entries were missing because those PRs predated the per-PR entry habit:

| Issue | What |
|---|---|
| #392 | `bicg` computing `rho` from two preconditioned vectors |
| #393 | `qmr` applying the preconditioner on both sides at half strength |
| #394 | `ilu_0` / `block_diagonal` `adjoint_solve` stubs |
| #401 | `multigrid.hpp` not self-contained |
| #386 | `long double` reference inverting assertions by platform |

The three Krylov defects are grouped under one heading, because they share a cause worth stating once rather than three times: every preconditioned `bicg`/`qmr` test used `pc::identity` — the single preconditioner for which a doubled, wrongly-sided, or wrongly-adjointed `M⁻¹` is indistinguishable from a correct one. That is why `test_preconditioner_adjoint.cpp` asserts the defining identity `⟨M⁻¹b, c⟩ == ⟨b, M⁻ᴴc⟩` instead of testing convergence.

## Note

Docs-only; no code touched, so no test results to report. Session logs are not in `sync-content.mjs`'s `FILE_MAP` — they are internal records rather than published pages, consistent with the existing ones.

## Summary by CodeRabbit

- **Bug Fixes**
  - Documented corrections improving the reliability of preconditioned Krylov solvers, including BiCG, QMR, ILU(0), block-diagonal, and SSOR operations.
  - Recorded fixes addressing multigrid portability and numeric testing consistency.
  - Updated benchmark and regression findings, including cache-block sizing and stale-build issues.

- **Documentation**
  - Added a detailed engineering session record covering diagnostics, measurements, resolved issues, open work, and testing lessons.


Ravenwater:
Five of six taken in 5f04d9d, one declined.

**[5] the order-independence claim — this one was a real error, and it reached beyond this PR.** I wrote that a sparse matvec over unsorted indices is "EXACTLY correct" because it sums "in storage order, which is order-independent". That is wrong. Reordering preserves the *set* of products so the sum is still right, but floating-point addition is not associative and a different order can round differently. Correct **to rounding**, not bit-identical.

The load-bearing point survives — `A*x` cannot *detect* unsorted indices either way, which is the whole reason `test_spgemm.cpp` asserts index structure instead of a residual — but the claim as written was false, and it was sitting in a **shipped header comment**. Fixed in all four places it appeared: `include/mtl/operation/spgemm.hpp`, `tests/unit/operation/test_spgemm.cpp`, the #402 CHANGELOG entry, and the session log (prose *and* the table row, which the review did not flag but was equally wrong).

**[1] BiCG↔CG qualification — correct, and I had already established it earlier in the day.** The reduction needs the *preconditioner* to be SPD as well; that is a statement about `M`, not about `A`, and it is exactly why the regression test uses `pc::diagonal` (symmetric whatever `A` is) rather than `ssor`. The CHANGELOG simply did not say so. Now does.

**[2] "half strength" — worse than unsupported, it is inverted.** `qmr.hpp` documents the old behaviour as the preconditioner applied **twice** and the split as `M₁ = M`, `M₂ = I`. "Half strength" implies a partial application on each side; the defect was a doubled one. Rewritten to match the code, including the `w̃ᵀM⁻¹ṽ` the algorithm actually needs.

**[4] MD040** and **[6] MD018** — both applied.

**[3] "Full day session" → "Full-day session" — declined.** The hyphen is correct English, but all **11** existing session logs use the unhyphenated form in that template field. Changing only the newest one makes it the odd one out; consistency in a template field is worth more than the hyphen. Happy to change all twelve in a separate pass if you would rather have it right everywhere.

164/164 on gcc — the code changes are comments only, but they are in headers, so worth confirming rather than assuming.

Issue #432, opened 2026-08-13
fix(simd): cache detection is core-dependent on hybrid CPUs and treats shared caches as private


Ravenwater:
## What happened

Running the #426 detection on the `i7-12700K` (Step 0 of #430) reported:

```
l1d=32768  assoc=8  l2=2097152  l3=26214400  line=64
runtime  mr=6 nr=4 kc=512 mc=256 nc=2048     (fp64)
default  mr=6 nr=4 kc=512 mc=32  nc=2048
```

`docs/benchmarks/systems.md` records that machine as **L1d 512 KiB across 12 instances, L2 12 MiB across 9 instances** — i.e. 8 P-cores at 48 KB L1d / 1.25 MB L2, plus 4 E-cores at 32 KB L1d sharing a **2 MB cluster L2**. So `l1d = 32 KB` and `l2 = 2 MB` are both **E-core** figures: the detecting thread was scheduled onto an E-core.

Two defects, and the second is the more serious.

## 1. Detection is non-deterministic on a hybrid part

`fill_caches_x86` issues CPUID leaf 4, which describes **the core the thread is currently running on**. Nothing pins it, so the same binary on the same machine yields different blocking parameters run to run:

| detected on | L1d | L2 | fp64 `kc` | fp64 `mc` |
|---|---|---|---|---|
| no detection (Haswell default) | 32 KB | 256 KB | 512 | 32 |
| E-core (`taskset -c 16`) | 32 KB, 8-way | 2 MB (shared by 4) | 512 | **256** |
| P-core (`taskset -c 0`) | 48 KB, 12-way | 1.25 MB | **768** | **106** |

`kc` held still only by coincidence — Gracemont's 32 KB L1d happens to equal the hardcoded default. On a P-core it moves too.

A performance parameter that depends on where the scheduler happened to put a thread during static initialization is not a parameter, and it makes any measurement on this machine unreproducible.

## 2. A shared cache is being used as a private one

The E-cluster L2 is **2 MB shared by 4 cores** — 512 KB per core. `derive_blocking` treats `l2_bytes` as private and sizes the packed A block to half of it, so `mc = 256` assumes 4x more L2 than a core actually has. The block overflows L2 on every core in the cluster.

[...]

Issue #430, opened 2026-08-13
bench(simd): multi-hardware experiment to choose the nc / jc_nt model


Ravenwater:
## Why this exists

#429 asks for a `balanced_nc` and then for the detected L3 to feed `nc`. Neither can be decided on the development machine: #426 established that the two defects it hit were both **invisible locally** — one needed a wide/short shape (square shapes make `jc_nt` structurally 1), the other needed an L1d other than 32 KB. A model chosen from one microarchitecture would be a repeat of the hardcoded Haswell `hw_traits` this whole line of work is removing.

So this issue is the record that **#429 is blocked on multi-hardware data**, and the design of the experiment that produces it.

## Hardware

Four machines spanning 4 microarchitectures, 4 SIMD widths, and an 8x range of L2:

| Machine | Micro-arch | Threads | ISA / Highway target | Registered |
|---|---|---|---|---|
| `i7-12700K` | Alder Lake (hybrid 8P+4E) | 20 (8 P-cores used) | AVX2, 4 doubles | [yes](docs/benchmarks/i7-12700k.md) |
| `ryzen-9-8945hs` | Zen 4 | 16 (8 physical used) | AVX-512, 8 doubles | [yes](docs/benchmarks/ryzen-9-8945hs.md) |
| Jetson (Orin-class) | ARM Cortex-A78AE | 8 | NEON, 2 doubles | **needs a systems.md entry** |
| `xeon-e5-2420v2` | Ivy Bridge-EP | 12 (6 physical) | SSE only — no AVX2/FMA, 2 doubles | **needs a systems.md entry** |

The Xeon earns its place precisely because it is the odd one out: no FMA, so the micro-kernel is compute-bound at ~9.3 GFLOP/s and cache effects cannot show. It is the **negative control machine** — a model that appears to help there is suspect.

## Step 0 — validate #426 on real hardware (do this first, it is cheap)

#426 merged with its central claim untested: this host's L1/L2 happen to equal the Haswell defaults, so `kc`/`mc` never moved. Before any modelling:

1. Run `util_test_cache_info` on each machine and record the detected hierarchy.
2. Record `derive_blocking<float/double>` output (`mr nr kc mc nc`) against `default_blocking`.
3. Confirm detection agrees with `lscpu` / sysfs / `sysctl` on each.

Deliverable: a table of detected-vs-default `kc`/`mc` per machine. **This is the first evidence that #426 does anything at all**, and it costs one test run per box.

## The models to discriminate

[...]

Issue #429, opened 2026-08-13
perf(simd): balance the jc partition (balanced_nc)


Ravenwater:
> **RESOLVED for `balanced_nc` (#490, `2ce0819`). The L3 half was FALSIFIED and
> removed from scope — see "The second half, and why it is gone" below.**
>
> The evidence is #479: six candidate `nc` models timed on four
> microarchitectures and two dtypes, 840 arms, every one verified bit-identical
> to the baseline. Method and the full hypothesis register:
> `docs/performance/benchmarking-methodology.md`.

## Context

Split out of #222 / #426, with a measured reproduction. #426 detects the cache hierarchy and feeds L1 -> `kc` and L2 -> `mc`, but **deliberately does not apply the detected L3**, because doing so regresses wide/short threaded GEMM. This issue covers making that safe.

## The mechanism

`l3_bytes` feeds only `nc`, and `nc` is the jc **block count**: `njb = ceil(n/nc)`. The threaded nest hands jc-blocks to teams round-robin (`gemm_blocked.hpp`):

```cpp
for (std::size_t jb = jc_id; jb < njb; jb += jc_nt) { ... }
```

So the critical path is `ceil(njb / jc_nt)`, and an `njb` that is not a multiple of `jc_nt` leaves teams unequal. Applying this machine's detected 15 MB L3 moved `nc` 2048 -> 3840 and `njb` 8 -> 5, so two teams took **3 and 2** blocks instead of 4 and 4.

This is exactly the failure mode of #408, one loop over: there a register-tile change moved `mc` 64 -> 60 and `nib` 16 -> 18, unbalancing the **ic** partition. The ic loop has `detail::balanced_mc` to absorb it. The jc loop has no equivalent.

## Reproduction

Xeon E5-2420 v2 (Ivy Bridge-EP, 6C/12T), Highway (128-bit target), fp64, `MTL5_NUM_THREADS=6`, `m=96 n=16384 k=1024`, interleaved, min of 9:

| nc | njb | teams | blocks/team | time |
|---|---|---|---|---|
| 2048 (compile-time default) | 8 | 2 | 4, 4 | **0.114 s** |
| 3840 (detected 15 MB L3) | 5 | 2 | 3, 2 | **0.126 s** (10-25% slower) |

[...]

PR #433, opened 2026-08-13
perf(simd)!: make cache detection opt-in — measured slower than the defaults — plus the A/B harness that showed it

Ravenwater:
## Summary

#426 wired the detected cache hierarchy into `kc`/`mc` on the assumption that real cache sizes must beat figures hardcoded to a Haswell-class core. **This PR measures that assumption and finds it false**, so detection becomes opt-in and shipped blocking returns byte-identical to pre-#426.

## The measurement

i7-12700K, Golden Cove, P-cores pinned, AVX2, fp64, 5 interleaved rounds, min of 5. `detected kc=384 mc=213` against `default kc=256 mc=64`. Ratio is detected/default throughput, so below 1 is a loss:

| shape | T=1 | T=8 |
|---|---|---|
| 1024³ | 0.965 | **0.551** |
| 2048³ | 0.920 | 0.929 |
| 4096³ | 0.901 | 0.985 |
| 852 × 8192 × 1024 | 0.914 | 0.806 |
| 852 × 12288 × 1024 | 0.906 | 0.675 |

**Nine losses, one tie, no wins.**

## Three causes, only one of them about cache sizing

1. **At T=1 the larger blocks are just slower** — ~9% at every size, no threading involved. The analytical "half of L1, half of L2" sizing is beaten by the hand-tuned constants on this microarchitecture. This is the finding that matters most, and no amount of fixing the thread nest repairs it.
2. **A larger `mc` starves the thread partition.** `gemm_blocked` fixes `ic_nt = min(budget, nib)` from the *unbalanced* block count, before `balanced_mc` runs. At 1024³, `mc` 64 → 213 takes `nib` 16 → 5, so **five threads of eight** do the work — which is the 0.551.
3. **A larger `kc` inflates the packed B panel.** Once the jc loop splits into teams each holds one: 2 × 12.6 MB against a 25 MiB L3.

Causes 2 and 3 are the #408 / #429 family — a cache-derived parameter moving a *block count* that the thread partition is acutely sensitive to. Both are defects that predate this work; detection merely exposed them.

## What changes

- **`MTL5_ENABLE_CACHE_DETECTION`** — detection is opt-in. Without it, `detected_hw_traits()` returns the compile-time defaults, so shipped blocking is byte-identical to pre-#426 on every machine.
- Everything else **stays**: the detection code, its CPUID/sysfs/sysctl paths, the tests, the harness, the i7 readings. #429, #430 and #432 all build on them, and the question is now *decidable* rather than assumed.
- Tests assert **both** configurations — with the macro, detected values must be applied; without it, `runtime_blocking` must equal `default_blocking` down to `kc` and `mc`. That unconditional assertion is what fails first if the default is ever flipped back without a measurement.

## The harness that produced this

- `benchmarks/bench_blocking_ab.cpp` built twice from one source; the arms differ *only* by the macro.
- `benchmarks/run_blocking_ab.sh` pins (`BENCH_PCPUS`, as `run_scaling.sh` does), interleaves within a session, min-of-N. **Shapes are derived from the machine**, once, from the detected arm and passed to both — the jc-parallel regime needs `nib ≤ T/2` and `njb ≥ 2`, and both bounds contain that machine's own `mc`/`nc`. Every #426 measurement was square, which is exactly where `jc_nt` is structurally 1 and none of this can appear.
- `benchmarks/analyze_blocking_ab.py` refuses to call a point unless the arms differ structurally, ranges don't overlap, and the effect clears 2%.

Validated first on the Xeon, a **null machine** (its L1/L2 already equal the defaults, so both arms compile identically and the right answer is "no difference"). The analyzer failed that on its first version — with 2 rounds it called a 4.8% "win" between two byte-identical configurations — and now detects identical parameters and declines to call anything. Structural control, not a statistical one.

## Docs

`docs/benchmarks/systems.md` gains both pinned cache readings for the i7 (it reports **two** hierarchies depending on where the detecting thread lands, #432) and the full A/B table above.

## Test Results

| | gcc | clang |
|---|---|---|
| full unit suite (detection off, shipped) | 164/164 | 164/164 |
| unit suite compiled **with** `MTL5_ENABLE_CACHE_DETECTION` | 63 assertions, 7 cases, pass | — |
| both A/B arms build + run | OK | — |

## Breaking change

Builds relying on #426's automatic `kc`/`mc` sizing must now define `MTL5_ENABLE_CACHE_DETECTION`. It was the default for less than a day and measured slower; no released version shipped it as the default.

Relates to #426, #429, #430, #432

### DIFF
--- a/benchmarks/CMakeLists.txt
+++ b/benchmarks/CMakeLists.txt
@@ -38,6 +38,18 @@ target_link_libraries(bench_sparse PRIVATE MTL5::mtl5)
 target_include_directories(bench_sparse PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/..)
 set_target_properties(bench_sparse PROPERTIES FOLDER "Benchmarks")
 
+# Cache-blocking A/B (#426 / #432): the SAME source built twice, once with
+# runtime cache detection and once with MTL5_DISABLE_CACHE_DETECTION, so the two
+# binaries differ only in kc/mc. Driven by benchmarks/run_blocking_ab.sh, which
+# pins and interleaves them. The `-default` arm is the pre-#426 behaviour.
+foreach(_arm detected default)
+    add_executable(bench_blocking_ab_${_arm} bench_blocking_ab.cpp)
+    target_link_libraries(bench_blocking_ab_${_arm} PRIVATE MTL5::mtl5)
+    target_include_directories(bench_blocking_ab_${_arm} PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/..)
+    set_target_properties(bench_blocking_ab_${_arm} PROPERTIES FOLDER "Benchmarks")
+endforeach()
+target_compile_definitions(bench_blocking_ab_default PRIVATE MTL5_DISABLE_CACHE_DETECTION)
+
 # Future: individual focused benchmarks can be added here
 # add_executable(bench_gemm bench_gemm.cpp)
 # target_link_libraries(bench_gemm PRIVATE MTL5::mtl5)
--- a/benchmarks/CMakeLists.txt
+++ b/benchmarks/CMakeLists.txt
@@ -38,17 +38,18 @@ target_link_libraries(bench_sparse PRIVATE MTL5::mtl5)
 target_include_directories(bench_sparse PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/..)
 set_target_properties(bench_sparse PROPERTIES FOLDER "Benchmarks")
 
-# Cache-blocking A/B (#426 / #432): the SAME source built twice, once with
-# runtime cache detection and once with MTL5_DISABLE_CACHE_DETECTION, so the two
-# binaries differ only in kc/mc. Driven by benchmarks/run_blocking_ab.sh, which
-# pins and interleaves them. The `-default` arm is the pre-#426 behaviour.
+# Cache-blocking A/B (#426 / #430 / #432): the SAME source built twice, once with
+# MTL5_ENABLE_CACHE_DETECTION and once without, so the two binaries differ only
+# in kc/mc. Driven by benchmarks/run_blocking_ab.sh, which pins and interleaves
+# them. The `-default` arm is what MTL5 ships; the `-detected` arm is the opt-in
+# under test, which measured a loss on the i7-12700K.
 foreach(_arm detected default)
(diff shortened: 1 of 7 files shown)


Ravenwater:
All seven fixed in 746780a. Every one was valid; two of them were errors in my own prose, and both flattered the finding.

**Integrity — a partial run must not produce confident verdicts.** The analyzer compared the *intersection* of the two CSVs and accepted unequal round counts through `min(len(a), len(b))`, so a half-finished arm could still yield decisive wins and losses. It now refuses outright and names the offending points:

```
INCOMPLETE A/B -- refusing to compare.
  only in detected: ('double', 128, 128, 128, 1)
  unequal rounds:   ('double', 64, 64, 64, 1) detected=2 default=1
```

Same class in the script: `THREADS` larger than `BENCH_PCPUS` was accepted because `cut` returns the whole shorter list when asked for more fields than it has — so the run used more workers than pinned cores while the CSV recorded the larger count. Now rejected, with non-integer counts.

**Arm order alternates between rounds.** Fair point: a fixed order folds warm-up, frequency ramp and thermal drift into the ratio in one direction. Odd rounds run detected first, even rounds default first.

**Overflow and option validation.** All confirmed reachable and all now rejected:

```
bad shape '0,10,10' (want m,n,k, all > 0)
shape 18446744073709551615,4,4 overflows size_t
--threads: needs at least one count
--reps must be > 0
--dtype must be double or float (got 'int')
```

The overflow one is the sharpest: a wrapped `m*k` allocates a short vector while `gemm_blocked` still walks the original extents — an out-of-bounds write, not a failed run. Shapes that are merely too large for RAM are left to fail as `bad_alloc`, which is a clear enough signal.

**The two accuracy findings were the most useful, because they were mine.**

"Slower by ~9% at every size" is not what the data says. The T=1 losses are 3.5% (1024³), 8.0%, 9.9%, 8.6%, 9.4% — so 1024³ is *a third* of what I claimed, and the round number came from eyeballing four of five points. Corrected to "3.5% at 1024³ and 8–10% at the rest" in both `blocking.hpp` and `systems.md`.

And `systems.md` said every ratio below 1 is a loss while the summary called `0.985` a tie — the analyzer's 2% floor was doing work the prose didn't mention. Now stated where the table is read.

Neither changes the conclusion (nine losses, one tie, no wins), but both would have made the record less trustworthy than the measurement it describes.

Verified: GCC 164/164, Clang 164/164; every new rejection path exercised; and the previously-collected Xeon data still analyzes cleanly through the new gate.

PR #434, opened 2026-08-13
bench(simd): commit the i7-12700K cache-blocking A/B data

Ravenwater:
## Summary

The raw measurements behind the verdict in `docs/benchmarks/systems.md` and the opt-in decision in #433, measured on `sw-21`.

`main` carried the summary table but not the data. With these committed, the conclusion is **reproducible**:

```
$ ./benchmarks/analyze_blocking_ab.py benchmarks/data/blocking_ab_{detected,default}.csv
blocking (double):  detected mr,nr,kc,mc,nc = ('6','8','384','213','4096')   default = ('6','8','256','64','4096')

       852x8192x  1024   1      68.04      62.19   0.914  DEFAULT faster
       852x8192x  1024   8     412.82     332.89   0.806  DEFAULT faster
      852x12288x  1024   1      67.58      61.21   0.906  DEFAULT faster
      852x12288x  1024   8     413.53     279.34   0.675  DEFAULT faster
      1024x1024x  1024   1      70.33      67.87   0.965  DEFAULT faster
      1024x1024x  1024   8     463.62     255.47   0.551  DEFAULT faster
      2048x2048x  2048   1      68.10      62.68   0.920  DEFAULT faster
      2048x2048x  2048   8     462.38     429.57   0.929  DEFAULT faster
      4096x4096x  4096   1      70.01      63.09   0.901  DEFAULT faster
      4096x4096x  4096   8     450.27     443.54   0.985  noise (<2%)

0 faster, 9 slower, 1 indistinguishable
```

That matches the published table exactly, and it passes the integrity gate added in #433 (equal key sets, equal round counts, matching checksums).

## Contents

50 rows per arm — 5 rounds x 2 thread counts x 5 shapes — plus a `.sysinfo` sidecar per arm recording the hierarchy that arm was blocked for. The sidecars are what make the rows interpretable later, and on this machine they also prove the run was pinned correctly:

```
cache_l1d_bytes=49152      <- P-core (Golden Cove), not the 32768 E-core value
cache_l2_bytes=1310720     <- private 1.25 MiB, not the 2 MiB E-cluster L2
cache_l3_bytes=26214400
```

Had the run drifted onto an E-core (#432), the sidecar would say so and the numbers would have to be discarded. That is the point of committing them.

## Note on the source branch

These files were committed on `bench/blocking-ab-harness`, which was cut **before** #433's opt-in flip and review hardening. That branch still carries `MTL5_DISABLE_CACHE_DETECTION` and the un-hardened analyzer, so merging it would revert both. Only the four data files are taken here; the branch itself should be deleted rather than merged.

`scripts/run_cache_detection.sh` from that branch is not carried over as a file. It was a personal runbook and had already gone stale — it named `MTL5_DISABLE_CACHE_DETECTION`, which #433 renamed when detection became opt-in. Its live content is folded into `systems.md` instead, next to the per-machine pinning policies the procedure depends on, so the instructions and the topology they assume cannot drift apart.

The stale branch has been deleted (all four data files verified byte-identical here first).

Relates to #426, #430, #432, #433

## Summary by CodeRabbit

* **Documentation**
  * Added system environment metadata to default and detected blocking benchmark results.
  * Benchmark records now include CPU, operating system, compiler, C++ standard, build type, and cache configuration details.
  * Added instructions for running cache-blocking A/B benchmarks on new machines, including CPU pinning, native ISA settings, machine-specific shapes, CSV preservation, and result interpretation.


Ravenwater:
Both fixed in 25d4e5d, and both were valid.

**The scope finding is the important one.** The section called itself a "cache-blocking A/B" without stating that the `detected` arm moves **only `kc` and `mc`**. Verified against the code: `with_detected_caches` applies L1, L1-assoc, line size and L2 — never `l3_bytes` — and `runtime_blocking` additionally pins `nc` to the compile-time value so it cannot drift indirectly through `kc`.

That gap could produce a wrong conclusion rather than just confusion. Most machines have an L3 far from the 8 MiB default (the Zen 4 next in line for this has 32 MiB), so someone could run the runbook, see no win, and record "cache detection does not help here" — when half the question was never in play. The section now says plainly that it answers *"do detected `kc`/`mc` beat the hand-tuned constants?"* and not *"does cache detection help?"*, with the reason L3 is excluded (`nc` sets the jc block count, the round-robin team assignment is sensitive to it, and applying detected L3 measured a 10–25% regression on a wide/short shape) and a pointer to #429 for when that changes.

**The placeholder finding I verified both directions**, because "not valid shell" is testable rather than a matter of taste:

```
$ bash old.sh
old.sh: line 1: one: No such file or directory
```

`BENCH_PCPUS=<one logical id per physical core>` *parses* — which is what makes it nasty — and then fails at runtime because `<one` is a redirect. The replacement is the i7's real P-core map with the Xeon's alongside and an explicit instruction to substitute from `lscpu -e=CPU,CORE`; extracted from the rendered page it now passes `bash -n`.

Docs re-validated: H1 intact for frontmatter, all 5 relative links resolve, no unescaped angle brackets outside code spans.

PR #436, opened 2026-08-13
fix(simd): make cache detection deterministic and shared-cache aware

### LINKED ISSUE #432: fix(simd): cache detection is core-dependent on hybrid CPUs and treats shared caches as private

Ravenwater:
## Summary

Resolves #432, both halves, so the #430 experiment can run on the Zen 4 and the Jetson without the detector contradicting itself. Detection is opt-in, so **shipped blocking is unaffected** — this is about the experiment being trustworthy.

## Determinism

CPUID leaf 4 describes whichever core the caller happens to be on, and nothing pinned it — so on the i7 the same binary reported a P-core or an E-core hierarchy run to run.

Linux now uses the **sysfs** walk on every ISA including x86, because sysfs is per-CPU data read by id rather than a property of the current core. It scans the cpus the process may actually run on (`sched_getaffinity`) and keeps, per level, the entry with the **smallest per-core budget**:

- **under `taskset`** that is the pinned set, so detection describes the cores the work will use — exactly what the harness needs;
- **unpinned on a hybrid machine** it is the whole machine, and taking the minimum means blocks fit whichever core the work lands on rather than overflowing the smaller kind.

Either way the answer is a function of the machine and the affinity mask, never of the scheduler. CPUID stays for Windows and as a fallback where sysfs is absent, documented as per-current-core.

## Sharing — counted in cores, not CPUs

`cache_info` reports `*_sharing_cores` beside each size and `with_detected_caches` divides by it, so a shared cache contributes only its per-core share.

**The core-vs-CPU distinction is the whole correctness of this.** sysfs reports this Xeon's L1d and L2 as `shared_cpu_list=0,6` — the two SMT siblings of core 0:

```
index0: level=1 type=Data     size=32K    shared_cpu_list=0,6
index2: level=2 type=Unified  size=256K   shared_cpu_list=0,6
index3: level=3 type=Unified  size=15360K shared_cpu_list=0-11
cpu0: core_id=0    cpu6: core_id=0
```

Counting *CPUs* would halve L1d and L2 on every hyperthreaded machine — while the pinning policy runs one thread per physical core and leaves the sibling idle. Counting *cores* gives 1 there (no division) and 4 for an Alder Lake E-cluster L2, which is the case that actually needs discounting. Identity is `(package, core_id)`, since `core_id` is only unique within a package.

## Verification

Xeon E5-2420 v2, 6C/12T, homogeneous — identical from every core and every affinity mask:

```
cpu0   l1d=32768/1 l2=262144/1 l3=15728640/6 | model l1=32768 l2=262144
cpu3   l1d=32768/1 l2=262144/1 l3=15728640/6 | model l1=32768 l2=262144
cpu6   l1d=32768/1 l2=262144/1 l3=15728640/6 | model l1=32768 l2=262144
cpu11  l1d=32768/1 l2=262144/1 l3=15728640/6 | model l1=32768 l2=262144
0-5    l1d=32768/1 l2=262144/1 l3=15728640/6 | model l1=32768 l2=262144
free   l1d=32768/1 l2=262144/1 l3=15728640/6 | model l1=32768 l2=262144
```

Both directions of the core-vs-CPU rule are exercised: the SMT pair collapses to **1**, and the six cores sharing L3 are counted as **6**.

- GCC **164/164**, Clang **164/164**
- Tests also pass compiled **with** `MTL5_ENABLE_CACHE_DETECTION` (76 assertions, 9 cases)
- The non-Linux CPUID fallback still compiles and returns correct values (forced that branch on this host)
- New tests pin the sharing rule directly: a 4-core cluster L2 divides, an SMT pair does not, unknown sharing (0) is treated as private rather than dividing by zero

## What this does not prove

**The hybrid case cannot be reproduced on any machine I can run.** The Xeon is homogeneous with private L1/L2, so it demonstrates determinism and the SMT rule but not the E-cluster discount in situ. On the i7 this should now report 48 KiB / 1.25 MiB pinned to P-cores, and the smaller E-cluster share unpinned — **worth confirming there before the Zen 4 and Jetson runs**:

```bash
taskset -c 0  ./build-ci/tests/unit/util_test_cache_info -s | grep -E "l1d=|runtime "
taskset -c 16 ./build-ci/tests/unit/util_test_cache_info -s | grep -E "l1d=|runtime "
```

Both should now be stable across repeats, and the unpinned run should equal one of them rather than varying.

Resolves #432
Relates to #430

## Summary by CodeRabbit

* **Improvements**
  * Improved CPU cache detection across systems with shared and multi-level caches.
  * Cache sizes are now normalized per physical core for more accurate hardware tuning.
  * Detection considers all CPUs available to the process and provides more consistent results.
  * Added support for reporting cache-sharing information for L1, L2, and L3 caches.

### DIFF
--- a/docs/benchmarks/systems.md
+++ b/docs/benchmarks/systems.md
@@ -45,24 +45,32 @@ This machine has **two different cache configurations in one socket**, and which
 one MTL5 detects depends on where the detecting thread is scheduled (#432). Both
 readings are recorded because neither is "the" answer for this CPU:
 
-| Pinned to | L1d | L2 | L3 | fp64 `kc` `mc` | fp32 `kc` `mc` |
+| Pinned to | L1d | L2 (raw / cores sharing) | L3 | fp64 `kc` `mc` | fp32 `kc` `mc` |
 |---|---|---|---|---|---|
-| P-core, `taskset -c 0` (Golden Cove) | 48 KiB, 12-way | 1.25 MiB **private** | 25 MiB | `768` `106` | `768` `213` |
-| E-core, `taskset -c 16` (Gracemont) | 32 KiB, 8-way | 2 MiB **shared by 4 cores** | 25 MiB | `512` `256` | `512` `512` |
+| P-core, `taskset -c 0` (Golden Cove) | 48 KiB, 12-way | 1.25 MiB / **1** | 25 MiB / 12 | `768` `106` | `768` `213` |
+| E-core, `taskset -c 16` (Gracemont) | 32 KiB, 8-way | 2 MiB / **4** | 25 MiB / 12 | `512` `64` | `512` `128` |
 | — Haswell defaults, for comparison | 32 KiB | 256 KiB | 8 MiB | `512` `32` | `512` `64` |
 
 Read with `util_test_cache_info -s` from a `ci`-preset build. Blocking figures are
 the 128-bit (SSE) build; an AVX2 build divides them by the wider SIMD width.
 
-Two things follow, and both matter when reading any result from this machine:
+Three things follow, and all matter when reading any result from this machine:
 
-1. **Unpinned detection is not reproducible here.** The same binary reports either
-   row depending on the scheduler. Every measurement on this machine must pin, and
-   must record which core class it pinned to.
-2. **The E-core L2 is shared by four cores**, so its 2 MiB supports about 512 KiB
-   per core — the `mc = 256` derived from it assumes roughly 4x the L2 a core
-   actually gets. The P-core's 1.25 MiB is genuinely private and needs no such
-   discount.
+1. **Detection is per-core-class and must be pinned.** The two rows are what the
+   detector reports under `taskset -c 0` and `taskset -c 16`; both are stable
+   across repeats since #432, but they are *different*, so every measurement must
+   record which class it pinned to. Unpinned, the detector reports the smallest
+   per-core budget across the cores the process may use — deterministic, and
+   chosen so blocks fit whichever kind the work lands on.
+2. **The E-core L2 is one 2 MiB instance shared by four cores**, so a core's share
+   is 512 KiB. The `mc` column reflects that discount: `64` rather than the `256`
+   the raw 2 MiB would give. Before #432 the raw figure was used, sizing the
+   packed A block for roughly 4x the L2 a core actually has.
+3. **Sharing is counted in physical cores, not logical CPUs.** The P-core L1d/L2
+   are shared by an SMT pair yet report `1`, because the pinning policy runs one
+   thread per physical core and leaves the sibling idle — counting CPUs would
+   halve both. The `25 MiB / 12` L3 is the same rule read the other way: 20
+   logical CPUs collapse to the 12 physical cores (8 P + 4 E) that share it.
 
 Under this page's pinning policy (P-cores, E-cores excluded) the relevant row is
 the first: against the Haswell defaults, detection blocks with a larger `kc` and
--- a/include/mtl/simd/blocking.hpp
+++ b/include/mtl/simd/blocking.hpp
@@ -212,11 +212,21 @@ inline constexpr blocking_params default_blocking = derive_blocking<T>(width<T>)
 /// a balanced_nc exists, L3 can be applied here. Tracked separately so that
 /// partition change gets its own measurement rather than riding along with cache
 /// detection.
+///
+/// A SHARED cache contributes only its per-core share. The BLIS model wants the
+/// L1/L2 one core gets to itself, and `cache_info` reports the instance size
+/// alongside how many physical cores share it, so the division happens here where
+/// the model is. It matters: an Alder Lake E-cluster publishes one 2 MiB L2 for
+/// four cores, and taking that at face value sized `mc` for roughly 4x the L2 a
+/// core actually has (#432). SMT siblings are deliberately NOT counted as sharers
+/// -- see cache_info -- so a hyperthreaded machine is unaffected.
 constexpr hw_traits with_detected_caches(hw_traits base, const util::cache_info& c) {
-    if (c.l1d_bytes  != 0) base.l1_bytes   = c.l1d_bytes;   // -> kc
+    const std::size_t l1_share = c.l1d_sharing_cores ? c.l1d_sharing_cores : 1;
+    const std::size_t l2_share = c.l2_sharing_cores  ? c.l2_sharing_cores  : 1;
+    if (c.l1d_bytes  != 0) base.l1_bytes   = c.l1d_bytes / l1_share;   // -> kc
     if (c.l1d_assoc  != 0) base.l1_assoc   = c.l1d_assoc;
     if (c.line_bytes != 0) base.line_bytes = c.line_bytes;
-    if (c.l2_bytes   != 0) base.l2_bytes   = c.l2_bytes;    // -> mc
+    if (c.l2_bytes   != 0) base.l2_bytes   = c.l2_bytes / l2_share;    // -> mc
     return base;
 }
 
--- a/include/mtl/util/cache_info.hpp
+++ b/include/mtl/util/cache_info.hpp
@@ -29,25 +29,44 @@
 
 #include <mtl/util/cpuid.hpp>
 
-#if !MTL5_HAS_X86_CPUID
-#  if defined(__APPLE__)
-#    include <sys/sysctl.h>
-#  elif !defined(_WIN32)
-#    include <cstdlib>
-#    include <fstream>
-#    include <string>
-#  endif
+#if defined(__linux__)
+#  include <cstdlib>
+#  include <fstream>
+#  include <sched.h>
+#  include <set>
+#  include <string>
+#  include <utility>
+#  include <vector>
+#elif defined(__APPLE__)
+#  include <sys/sysctl.h>
 #endif
 
 namespace mtl::util {
 
 /// Detected data-cache hierarchy, in bytes. 0 = not detected.
+///
+/// `*_bytes` is the cache's TOTAL size as the hardware reports it, and
+/// `*_sharing_cores` is how many distinct PHYSICAL cores share that instance, so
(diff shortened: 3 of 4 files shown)


Ravenwater:
Three of four fixed in 902c6ed; one declined with reasons.

**`cpu_set_t` sizing — fixed, and it was worse than it looked.** `cpu_set_t` is fixed at `CPU_SETSIZE` (1024), so on a host configured with more, `sched_getaffinity` fails with `EINVAL` and the code fell back to cpu0 — which need not be in the affinity mask at all. Detection would then describe a core the work can never run on, *silently*. That is a worse failure than the non-determinism this PR exists to fix. `allowed_cpus()` now grows a `CPU_ALLOC`'d set until it fits and only falls back when the call fails for some reason other than being too small; the fixed-size path is kept for a libc without `CPU_ALLOC`.

**Unknown sharing stored as 1 — fixed.** Storing `1` for an unreadable topology reports "known to be a private cache" about a cache whose topology we could not read, contradicting the struct's own documented contract that 0 means unknown. The stored value is now the raw count; only the local per-core divisor coerces it to 1, and `with_detected_caches` already treats 0 as private where it spends the budget.

**`hardware_concurrency()` bound — fixed, and you were right that it passed by luck.** I checked: this glibc's `hardware_concurrency()` is affinity-unaware, returning 12 both free and under `taskset -c 0,1`, so the assertion held on this toolchain while comparing a sysfs *physical-core* count against a hint with no defined topology scope. Replaced with an invariant of the topology itself — sharing widens monotonically with level, `l1d <= l2 <= l3` — which needs no external bound and holds on any machine.

---

**`std::filesystem::path` for the sysfs paths — declining, with reasons.**

The guideline it cites exists to stop hardcoded paths that break across platforms. These are not that: they are Linux kernel ABI paths (`/sys/devices/system/cpu/...`) inside `#if defined(__linux__)`, in a block that cannot compile anywhere else. There is no separator to get wrong and no platform for it to be wrong on.

Against that, `<filesystem>` is a heavy include, and this header is reachable from the core (`cache_info.hpp` → `simd/blocking.hpp` → gemm), so every translation unit that touches a matrix would pay for it. `<filesystem>` is currently confined to `io/png.hpp`, `io/spy.hpp` and `generators/testsuite.hpp` — all leaf headers, none in the core path — and I would rather not be the change that pulls it in. There is also direct precedent two files away: `system_info.hpp` opens `/proc/cpuinfo` and `/etc/os-release` as plain string literals.

Happy to revisit if you want the guideline applied literally regardless of the platform guard — it is a small change — but I do not think it improves this code.

Verified after the fixes: identical detection from cpu0, cpu6 and unpinned (`l1d=32768/1 l2=262144/1 l3=15728640/6`); GCC 164/164, Clang 164/164.

PR #441, opened 2026-08-14
perf(simd): plan the GEMM thread grid instead of greedily filling it

Ravenwater:
## Summary

Fixes the thread-grid half of #429. Two defects, both found by measuring three machines in #430, both costing **whole cores** rather than percentages. `plan_gemm_grid` is a **pure function**, so the decision is unit-testable and enumerable offline rather than inferred from a throughput change — which #430 also asked for.

## Defect 1 — `mc` was the cache bound only, so the partition could starve

`nib = ceil(m/mc)` could come out smaller than the thread budget, and `ic_nt = min(budget, nib)` then capped the grid permanently. On a **square** problem `njb == 1`, so no jc parallelism can make up the difference and the machine idles:

| machine | `mc` | `nib` | grid | measured |
|---|---|---|---|---|
| i7-12700K, 1024³ | 213 | 5 | **5 of 8** | 0.551 |
| Zen 4, 1024³ | 256 | 4 | **4 of 8** | 0.590 |

`mc` is now also bounded by `ceil(m/budget)` — enough blocks to give every thread one — floored at a register tile. It only ever *lowers* `mc`, so the L2 bound still holds, and it **binds only when the partition would starve**: at the shipped `mc = 64` with `m = 1024` on 8 threads the cap is 128 and nothing moves.

## Defect 2 — the factorization was greedy

Fill ic, then hand jc the integer quotient, discarding the remainder. At budget 6 with `nib = 4` it took `4 × 1` and left two cores idle while `3 × 2` and `2 × 3` were both legal (Jetson Orin Nano, 0.760). The search is now exhaustive over `ic` and maximises `ic_nt × jc_nt`, preferring the larger `ic_nt` on ties.

## The tie-break is not cosmetic — and it is what this host could measure

On the Xeon at `m=96 n=4096 k=1024`, 6 threads, the grid moves from `3 × 2` to `6 × 1` — **the same six threads** — but the live packed-B footprint halves:

```
before: mc=32 nib=3 grid=3x2=6   packed-B live = 2 x 8.4 MB = 16.8 MB   (L3 = 15.7 MB)
after : mc=16 nib=6 grid=6x1=6   packed-B live = 1 x 8.4 MB =  8.4 MB   (fits)
```

**20.97 → 30.64 GFLOP/s, 1.46×**, best of five interleaved runs. Shapes that were already balanced are unchanged — 1.00× at `m=192`, 1.03× at 1024³, both inside run-to-run spread.

Being precise about what this host proves: the **starvation** half cannot be demonstrated here. It needs `nib < budget` with `njb == 1`, and this machine's `mc = 32` at `m = 1024` gives `nib = 32` against 6 threads. Its evidence is the i7 and Zen 4 runs, and the unit tests encode those exact cases so they are pinned rather than remembered.

## Correctness

Results stay **bit-identical**: `mc` and the grid decide which rows are grouped and which thread owns a block, while the FMA order for a given C element is fixed by the untouched `pc` loop. Verified by identical checksums across the change and by the existing #297 determinism tests.

A degenerate-input sweep in the new tests caught a real contract violation on the way — `plan_gemm_grid` returned `mc = 0` for `mc_cache = 0`, and `mc` is a loop step.

## Test Results

| | gcc | clang |
|---|---|---|
| full unit suite | **165/165** | **165/165** |
| `detail_test_gemm_grid` (new) | 1928 assertions, 6 cases | — |
| `op_test_gemm_mt` (bit-identity) | pass | pass |

## What this does not do

It does not touch `nc` or apply the detected L3 — that is the rest of #429, and it is now clearly downstream: with the grid no longer wasting cores, the multi-thread losses that dominated the #430 data should shrink or vanish, and the cache model can finally be measured on its own. **Re-running the three machines after this lands is the natural next step**, and may well revisit #426's opt-in decision.

Relates to #429, #430

## Summary by CodeRabbit

* **Performance Improvements**
  * Improved threaded matrix multiplication workload distribution for better CPU utilization.
  * Dynamically adapts block sizes and parallel partitioning to available processing capacity and matrix dimensions.
  * Reduces thread starvation, especially for smaller or unevenly sized workloads.

* **Bug Fixes**
  * Improved handling of edge cases and constrained workloads while preserving existing blocking behavior.
  * Added regression coverage for parallel partitioning limits and degenerate inputs.

### DIFF
--- a/include/mtl/detail/gemm_blocked.hpp
+++ b/include/mtl/detail/gemm_blocked.hpp
@@ -102,6 +102,75 @@ inline std::size_t balanced_mc(std::size_t m, std::size_t mc_max, unsigned ic_nt
     return mc == 0 ? mc_max : mc;
 }
 
+/// The threaded nest's shape decision: the ic block size and the ic x jc thread
+/// grid. Pure, so it can be unit-tested and enumerated offline (#430) instead of
+/// being inferred from a throughput change.
+struct gemm_grid {
+    std::size_t mc;      ///< ic block size actually used (<= the cache bound)
+    std::size_t nib;     ///< ic blocks that mc produces
+    std::size_t njb;     ///< jc blocks
+    unsigned    ic_nt;   ///< threads across ic
+    unsigned    jc_nt;   ///< teams across jc
+};
+
+/// Choose mc and the grid for an m x n problem on `budget` threads.
+///
+/// Two defects this fixes, both found by measuring three machines (#430) and both
+/// costing whole cores rather than percentages:
+///
+/// 1. mc WAS THE CACHE BOUND ONLY, so `nib = ceil(m/mc)` could be smaller than
+///    the budget and cap ic_nt permanently -- with njb == 1 (square problems,
+///    where n <= nc) there is no jc parallelism to make up the difference and the
+///    machine simply idles. Measured: an i7-12700K at 1024^3 ran 5 of 8 threads
+///    (ratio 0.551 against the smaller-mc arm) and a Zen 4 ran 4 of 8 (0.590).
+///    mc is therefore also bounded by ceil(m/budget) -- enough blocks to hand
+///    every thread one -- floored at a single register tile. This only ever
+///    LOWERS mc, so the L2 bound is still respected, and it binds ONLY when the
+///    partition would otherwise starve: at the shipped mc = 64 with m = 1024 and
+///    8 threads the cap is 128 and nothing changes.
+///
+/// 2. THE FACTORIZATION WAS GREEDY: fill ic first, then give jc the integer
+///    quotient. That throws the remainder away -- at budget 6 with nib = 4 it
+///    picked 4 x 1 = 4 and left two cores idle, while 3 x 2 and 2 x 3 were both
+///    legal. Measured on a Jetson Orin Nano: 0.760 where the grid was the only
+///    difference. The search below is exhaustive over ic (a handful of values)
+///    and maximises ic_nt * jc_nt, preferring the larger ic_nt on ties because a
+///    wider ic team keeps the shared packed-B panel resident.
+constexpr gemm_grid plan_gemm_grid(std::size_t m, std::size_t n,
+                                   std::size_t mc_cache, std::size_t nc,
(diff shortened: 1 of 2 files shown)


Ravenwater:
Fixed in 38be189 — and checking it found the problem was **worse than reported**.

You flagged that `m=400` doesn't require the search. It doesn't, and neither does the Jetson case I wrote directly above it. **Neither of my two "search" tests exercised the search**, because wherever the `mc` cap can reach the budget it produces `nib >= budget`, and a greedy ic-first split lands on the same grid:

```
m=400 mc=100 B=6   ->  mc=67 nib=6   planner 6x1=6   greedy 6x1=6
m=64  mc=16  B=6   ->  mc=11 nib=6   planner 6x1=6   greedy 6x1=6   (jetson)
m=24  mc=100 B=6   ->  mc=6  nib=4   planner 3x2=6   greedy 4x1=4   <-- required
```

Deleting the exhaustive search would have left the entire suite green. That is the same "a test that exercises nothing" shape this project keeps running into — this time in a test I wrote specifically to prevent it.

Your suggested case is the right one, and the reason it works is worth stating: the **register-tile floor** forces it. At `m = 24` the cap wants `ceil(24/6) = 4`, below `mr`, so `mc` stops at 6 and `nib = 4 < budget`. Only the search reaches 6, via `3 × 2`.

Two tests now cover it, with the greedy split kept in the test file as a reference implementation so neither can go vacuous again:

- the `m = 24` case asserts the planner beats greedy on that shape, by value;
- a property sweep asserts `planner >= greedy` across 336 `(m, n, mc, budget)` combinations, so a regression on an unenumerated shape fails too.

**Verified negatively**, which is what the old tests could not do — reverting `plan_gemm_grid` to the greedy split:

```
tests/unit/detail/test_gemm_grid.cpp:93: FAILED:
  REQUIRE( g.ic_nt * g.jc_nt == 6 )
with expansion:
  4 == 6
```

The Jetson case keeps its assertions but its comment now says plainly that the **cap**, not the search, is what repairs it — so the next reader isn't misled about which defect that shape demonstrates.

GCC 165/165, Clang 165/165.

Issue #442, opened 2026-08-14
bench: a measurement contract — preflight checks and provenance in every result


Ravenwater:
## Problem

There is no measurement contract. `run_blocking_ab` is hardened; nothing else is, and even it cannot answer *"what code and what machine state produced this CSV?"*

Every guard it does have was added **after** the corresponding error had already produced a wrong result:

| error | guard | added after |
|---|---|---|
| another machine's data overwritten | `OUTDIR` required, per-machine | a Zen 4 run destroyed the i7's committed CSVs |
| identical configs reported as a win | null-run structural control | the analyzer called a 4.8% "win" between byte-identical arms |
| partial / unequal-round runs compared | integrity gate | — |
| sub-2% noise called a result | noise floor + non-overlapping ranges | — |
| arms computing different results | checksum equality, hard error | — |
| order effects (warm-up, thermal drift) | arm order alternates per round | review |
| clamped thread count invisible | `pool` column | Jetson requested 8 on a 6-core part; symptom was a speedup that "looked low" |
| wrong ISA measured silently | `-Arch` echoed, MSVC no-op documented | `MTL5_NATIVE_ARCH` is a no-op under MSVC |

The pattern is the point: each was reactive. The gaps below are the ones still open, and a preflight/provenance layer is the change that would catch the *next* error rather than documenting it afterwards.

## Gaps

**1. No provenance.** The `.sysinfo` sidecar records CPU, OS, compiler, caches — and **not** the git commit, working-tree state, or compile flags. Every one of the 18 sidecars committed today is untraceable to the code that produced it:

```
cpu_brand cpu_vendor cpu_arch cpu_logical_cores cpu_simd
os_name os_version compiler compiler_version cpp_standard build_type
cache_l1d_bytes cache_l1d_assoc cache_l2_bytes cache_l3_bytes cache_line_bytes
```

Nothing there distinguishes a run built with `/arch:AVX512` from one without — which matters, because the Zen 4 A/B turned out to be an AVX2 run when AVX-512 was intended, and the CSV cannot say so.

**2. No environment capture.** Governor, turbo/boost state, power mode, whether `jetson_clocks` ran — none recorded. The Jetson data was taken at 15 W with `schedutil` and unpinned clocks; that is written in `systems.md` prose because I asked, not because anything captured it. A reader of the CSV alone cannot know.

**3. No thermal record.** A throttled run produces low numbers indistinguishable from a slow configuration. There is no before/after temperature in the data.

[...]

PR #443, opened 2026-08-14
feat(bench): record build provenance in every benchmark sidecar

Ravenwater:
## Summary

Item (b) of #442. A committed CSV is evidence, and every sidecar written so far recorded the CPU, OS, compiler and caches — and **nothing identifying the code or the build**. Nothing in the Zen 4 data could say whether it was compiled with `/arch:AVX512` or without, which is exactly the question that turned out to matter for that run.

Sidecars now carry:

```
git_commit=e264a010b187    git_dirty=0
cxx_flags=<CMAKE_CXX_FLAGS as configured>
cmake_build_type=Release   pool_size=<effective threads>
detection_enabled=0|1
cache_l1d_sharing_cores=1  cache_l2_sharing_cores=1  cache_l3_sharing_cores=6
```

`cxx_flags` is the field that would have caught a real error *at the moment it was made* rather than days later — it is where the ISA selection lives, and the SIMD width it implies is what every blocking parameter divides by. `git_dirty` is **recorded, not prevented**: a result from a dirty tree is not reproducible, and saying so is more useful than pretending. The sharing counts close the loop with #432, so a reader can see the per-core budget the run was blocked for.

## Build time, not configure time — and why it matters

A configure-time commit goes stale the moment you commit and rebuild without reconfiguring, and a sidecar confidently naming the **wrong** commit is worse than one naming none. `cmake/GitInfo.cmake` reruns per build.

Verified end to end. Before committing this work, with a dirty tree:

```
git_commit=935ef118efbd
git_dirty=1
```

then `git commit` and a rebuild **with no reconfigure**:

```
git_commit=e264a010b187
git_dirty=0
```

A configure-time capture would still be reporting `935ef118efbd` / `dirty=1`.

And it costs nothing: `configure_file` leaves the header untouched when the rendered content is identical, so a second build with no git change re-runs the script and recompiles **nothing** (no `Building CXX` line).

## Scoping choice worth reviewing

Deliberately **not** on the `mtl5` INTERFACE. A per-commit define reaching every consumer would invalidate their objects on each commit — a full rebuild every time you commit. Benchmark targets depend on `mtl5_build_info` explicitly, and the header is also generated once at configure time so anything that includes it still compiles without taking that dependency.

## Not in this change

The environment half of #442 — governor, power mode, thermal headroom, concurrent-load detection — needs a preflight script per platform and is separate. So is bringing `run_scaling` and `run_sweeps` up to the contract.

## Test Results

| | gcc | clang |
|---|---|---|
| full unit suite | **165/165** | **165/165** |
| build-time regeneration | verified (SHA + dirty flag track the tree) | — |
| no spurious rebuild | verified (unchanged content → no recompile) | — |

Relates to #442, #430

## Summary by CodeRabbit

* **New Features**
  * Builds now record provenance metadata, including the source revision, working-tree status, compiler flags, and build type.
  * Build metadata is available to applications and tools through generated build information.
  * Cache-blocking benchmark reports now include build provenance, thread-pool size, cache-sharing details, and cache-detection status.

* **Bug Fixes**
  * Build metadata is generated consistently during configuration and builds, including when Git information is unavailable.
  * Build settings containing special characters are handled safely in generated metadata.


Ravenwater:
All three fixed in 23f1853. Each would have produced provenance that is **wrong** rather than absent — which is the failure this feature exists to prevent, so they were the right things to catch.

**Untracked files now count as dirty.** Not hypothetical in this repo: `tests/unit/CMakeLists.txt` uses `file(GLOB "*.cpp")`, so an untracked test source **is compiled into the build** while nothing tracked changed. The sidecar would have claimed the binary corresponds to a commit it does not.

```
tests/unit/CMakeLists.txt:4:  file(GLOB CONCEPT_SRC "concepts/*.cpp")
```

Switched to `--untracked-files=normal`. `.gitignore` still applies so build trees don't trip it, but a stray file in the working tree now sets the flag — which is the right incentive before a benchmark run. (It fires on this machine today, because of an untracked `trouble.txt` in the root. That is the mechanism working, not noise to suppress.)

**Values are escaped for the C++ string literal.** A Windows flag string breaks the header twice over — `/I C:\foo\bar /D S="q"` gives `\f` and `\b` as bogus escapes *and* terminates the literal early. Verified by feeding exactly that through the script:

```
#define MTL5_BUILD_CXX_FLAGS   "/I C:\\foo\\bar /D S=\"q\" -Wall"
$ ./esc_check
/I C:\foo\bar /D S="q" -Wall
```

Compiles, and round-trips byte for byte. Backslashes are escaped before quotes, or the escaping escapes itself.

**Multi-config generators record the real config.** `CMAKE_BUILD_TYPE` is empty under Visual Studio and Ninja Multi-Config, so `cmake_build_type` would have been blank on **exactly the machine where it matters** — the Zen 4 is the MSVC host, and it is the machine whose build configuration is already under suspicion. The build-time invocation now passes `$<CONFIG>`; the configure-time fallback records `unknown-at-configure-time` rather than a blank that reads like a missing field.

GCC 165/165, Clang 165/165, and the end-to-end sidecar re-verified after the changes.

Ravenwater:
All three fixed in a162b0d.

**Newlines are now encoded, not discarded.** You're right that dropping `\r` and folding `\n` to a space keeps the header *valid* while silently changing the value — wrong provenance rather than missing provenance, which is the exact failure this feature exists to prevent. Both are emitted as C escapes now, so the value survives intact.

**`build_git_dirty`'s "unknown" state is documented.** The comment promised only `"1"` or `"0"`, but it stays `"unknown"` when git is absent or the status query fails. The header now says so — and says that unknown is not clean, which is the part a reader would otherwise assume wrongly.

**The regression test now exists.** There was none; the escaping had only ever been checked by hand, which is not a defensible state for the thing that guarantees a benchmark's traceability. `tools/cmake/test_provenance_escaping.cmake` feeds a value carrying every hazard at once — Windows backslashes, an embedded quote, CRLF and a bare newline — through `GitInfo.cmake`, compiles the generated header, runs it, and requires the value back **byte for byte**. Registered as the `provenance_escaping` ctest.

Writing it surfaced a trap worth recording, because it is the same shape as the bug being fixed. My first version compared the probe's **stdout**, and `execute_process` strips carriage returns from captured output:

```
$ ./crlf | xxd        ->  410d 0a42     (4 bytes, CR present)
execute_process       ->  captured length=3, first_CR_at=-1
```

So it failed a *correct* escaping — and would equally have **passed** an escaping that dropped CR. It was measuring CMake, not the code. It now writes the macro to a file in binary mode and compares hex.

Verified negatively in both directions, which is what the hand-checking could never do: reverting the newline encoding fails the test, and removing the quote escaping fails it too.

GCC 166/166, Clang 166/166.

PR #444, opened 2026-08-14
feat(bench): preflight gates and machine-state capture for every measurement

Ravenwater:
## Summary

The first guard in this repo's benchmark harnesses written **before** the error rather than after it. `benchmarks/preflight.{sh,ps1}` runs ahead of any measurement, **fails** on conditions that invalidate a result, and emits `key=value` machine state into the sidecar either way.

Closes the first two acceptance items of #442; `run_scaling` / `run_sweeps` parity and the pre-contract labelling follow next.

## What it gates

| Condition | Action | Override |
|---|---|---|
| Working tree dirty | fail | `ALLOW_DIRTY=1`, recorded as `tree_git_dirty=1` |
| A build or benchmark already running | fail | none |
| Threads requested > available cpus | fail | none |
| Thermal headroom below margin | fail **only where a sensor and a limit are both readable** | `MIN_THERMAL_HEADROOM_C` (15) |
| Governor not `performance` | warn, record | — |

The thermal rule is asymmetric deliberately. Windows desktop firmware usually does not implement `MSAcpi_ThermalZoneTemperature`; failing there would make the Zen 4 machine unbenchmarkable, and passing silently would hide the gap. It records `unavailable` out loud. Governor is recorded rather than enforced because **all three** machines run a non-performance governor today — a hard gate would block every one of them.

## `build_isa` — the gap this work exposed

The stated purpose of the provenance epic was that *nothing distinguished a `/arch:AVX512` build from an AVX2 one*. Testing the pipeline end to end showed `cxx_flags` does **not** answer that:

- `MTL5_NATIVE_ARCH` adds `-march=native` via `target_compile_options` — it appears in no `CMAKE_CXX_FLAGS` variable
- the optimisation level lives in `CMAKE_CXX_FLAGS_<CONFIG>`

so a `-O3 -march=native` build recorded `cxx_flags=`, empty — the exact "empty value reads as a recorded fact" failure. Fixed two ways: per-config flags folded in (and the argument **quoted** — unquoted, any flag string containing a space would have split into separate `cmake` arguments), and `mtl::util::build_isa_list()` reads the compiler's own predefined macros.

```
default build:      build_isa=SSE2       cpu_simd=SSE2 AVX
-march=native:      build_isa=SSE2 AVX   cpu_simd=SSE2 AVX
```

Flags are intent; that is effect.

## Stale binaries

The binary records the commit it was **built** from; preflight records where the tree **is**. The harness compares them and writes `binary_stale`. Not hypothetical — the first end-to-end run of this change measured a binary two commits behind, and nothing in the numbers showed it.

## Tests, and what each can actually catch

- **`bench_preflight_contract`** runs whichever preflight applies to the platform and holds it to a **shared key list**, so each CI platform proves its own half. Two implementations of one contract running on disjoint machines drift silently otherwise — and PowerShell bugs reached `main` twice (#437/#438) because nothing in CI ever executed that file. Verified negatively: dropping a key fails it, and so does emitting a key with an **empty** value.
- **`build_isa`** is checked against the **runtime** CPU features. That is not a restatement of the implementation: a binary cannot use an ISA its CPU lacks, so a mis-mapping fails on every machine missing the feature.

Gates verified individually on the Xeon — dirty tree, thread budget, affinity-aware count under `taskset`, thermal margin, and a real competing process. Each fires; a clean run passes. Matching on *command lines* caught the shell that had merely typed the benchmark's path, so it matches process **names**.

## Changes

| File | What |
|---|---|
| `benchmarks/preflight.sh` / `.ps1` | new — gates and machine-state capture |
| `benchmarks/run_blocking_ab.sh` / `.ps1` | call preflight, append state, detect stale binaries |
| `include/mtl/util/system_info.hpp` | `build_isa_list()`, emitted as `build_isa=` |
| `CMakeLists.txt` | fold in per-config flags; quote the argument |
| `include/mtl/build_info.hpp.in` | `cxx_flags` documented as intent, pointing at `build_isa` |
| `tools/cmake/test_preflight.cmake`, `tests/CMakeLists.txt` | the contract test |
| `tests/unit/util/test_system_info.cpp` | `build_isa` ⊆ runtime features |
| `docs/benchmarks/systems.md` | "The run contract", documented once |

## Test Results

| Suite | Result |
|---|---|
| GCC | 167/167 |
| Clang | 167/167 |
| Preflight gates (manual, Xeon) | all five fire; clean run passes |
| End-to-end A/B run | sidecar carries all three provenance sources |

## Test plan

- [ ] Tier 1 CI on 8 platforms — Windows exercises `preflight.ps1` for the first time
- [ ] Promote when green

Relates to #442

## Summary by CodeRabbit

* **New Features**
  * Benchmark runs now capture system conditions, compiler-enabled ISA features, build provenance, thermal state, and repository status.
  * Added preflight checks for competing workloads, thread limits, thermal headroom, CPU availability, and power settings.
  * Benchmark reports warn when binaries may not match the current source revision.

* **Bug Fixes**
  * Build records now accurately reflect configuration-specific compiler options and enabled instruction sets.

* **Documentation**
  * Added guidance on benchmark run requirements, provenance data, thermal handling, and legacy reports.

* **Tests**
  * Added automated validation for preflight behavior and ISA reporting.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -79,12 +79,25 @@ configure_file(
 # content changes, so nothing rebuilds spuriously. Not part of the mtl5 INTERFACE
 # -- putting a per-commit define on every consumer would invalidate their objects
 # on each commit; benchmark targets depend on this target explicitly.
+#
+# CMAKE_CXX_FLAGS alone is usually EMPTY: the optimisation level lives in
+# CMAKE_CXX_FLAGS_<CONFIG> and MTL5_NATIVE_ARCH adds -march=native through
+# target_compile_options. A sidecar reading `cxx_flags=` on a -O3 -march=native
+# build records nothing while looking like it recorded something, so the
+# per-config flags are folded in here. The authoritative answer to "which ISA"
+# is build_isa in the sidecar, taken from the compiler's own predefined macros
+# (mtl/util/system_info.hpp) -- flags are the intent, that is the effect.
+set(MTL5_PROVENANCE_FLAGS "${CMAKE_CXX_FLAGS} $<$<CONFIG:Debug>:${CMAKE_CXX_FLAGS_DEBUG}>$<$<CONFIG:Release>:${CMAKE_CXX_FLAGS_RELEASE}>$<$<CONFIG:RelWithDebInfo>:${CMAKE_CXX_FLAGS_RELWITHDEBINFO}>$<$<CONFIG:MinSizeRel>:${CMAKE_CXX_FLAGS_MINSIZEREL}>")
+if(MTL5_NATIVE_ARCH)
+    set(MTL5_PROVENANCE_FLAGS "${MTL5_PROVENANCE_FLAGS} [MTL5_NATIVE_ARCH=ON]")
+endif()
+
 add_custom_target(mtl5_build_info ALL
     COMMAND ${CMAKE_COMMAND}
             -DSRC_DIR=${CMAKE_CURRENT_SOURCE_DIR}
             -DTEMPLATE=${CMAKE_CURRENT_SOURCE_DIR}/include/mtl/build_info.hpp.in
             -DOUTPUT=${CMAKE_CURRENT_BINARY_DIR}/include/mtl/build_info.hpp
-            -DCXX_FLAGS=${CMAKE_CXX_FLAGS}
+            "-DCXX_FLAGS=${MTL5_PROVENANCE_FLAGS}"
             -DCMAKE_TYPE=$<CONFIG>
             -P ${CMAKE_CURRENT_SOURCE_DIR}/cmake/GitInfo.cmake
     BYPRODUCTS ${CMAKE_CURRENT_BINARY_DIR}/include/mtl/build_info.hpp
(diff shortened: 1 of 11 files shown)


Ravenwater:
Four of five fixed in `adf49da`; the fifth I'm declining, with reasons.

**Argument parsing — correct, and worse than reported.** A trailing `--threads` didn't just fail to progress, it **spun forever**. Reproduced against the old shape:

```
$ timeout 5 bash oldparse.sh --threads
LOOPED 6 times without progress
```

And the non-numeric case was the more dangerous of the two: it exited **0** and silently skipped the thread-budget gate — a gate that looks enforced and isn't, which is the failure mode this whole script exists to prevent. Values are now demanded before the shift and validated as non-negative integers.

```
--threads            -> preflight: --threads requires a value           (rc=2)
--threads abc        -> must be a non-negative integer, got 'abc'       (rc=2)
--threads -4         -> must be a non-negative integer, got '-4'        (rc=2)
--threads 4          -> cpu_affinity=12, gates run
```

**macOS — correct.** No `nproc` there, so `cpu_affinity=0` and every run with `--threads N > 0` failed with "0 cpus available". A gate that blocks a platform outright isn't a stricter gate, it's a broken one. Now falls back to `sysctl`, verified against a stubbed macOS-shaped `PATH`:

```
nproc absent, sysctl present -> cpu_online=10  cpu_affinity=8  rc=0 (8 threads accepted)
```

**Postflight failure — accepted in substance, not in remedy.** You're right that the sidecar must never quietly omit `thermal_after_c`. But exiting non-zero discards a *completed* measurement session because a thermometer read failed: the numbers are already on disk and are unaffected by a missing probe, so aborting destroys good data to punish the wrong thing. Instead the failure is now recorded as `thermal_after_c=unavailable` and said out loud, in both harnesses — plus a second guard for a preflight that exits 0 while printing nothing:

```bash
case "$AFTER_KV" in
    *thermal_after_c=*) ;;
    *) AFTER_KV="thermal_after_c=unavailable" ;;
esac
```

`unavailable` is a fact; an absent key is a reader's guess. That distinction is the same one the empty-value check in the contract test enforces.

**Docs scope — correct.** "Every harness must satisfy this" was meant as a requirement, but it reads as a description. Added a per-harness status table with `run_scaling` and `run_sweeps` marked **not gated**, so nobody reads one of their CSVs as though it had been. They produced the published i7 and Ryzen numbers, which makes that ambiguity expensive.

---

**Declining: replacing the host-CPU comparison in the `build_isa` test.**

The suggested alternative — assert the expected tokens are present and serialize correctly — is a restatement of the implementation. It derives the expectation from the same macros the code reads, so it would pass *any* mis-mapping the code makes, including the one that matters (claiming `AVX512F` for an `/arch:AVX2` build). I deliberately avoided that shape after shipping vacuous tests on #441.

The host comparison is the only check here with an **independent** source of truth: a binary cannot execute an ISA its CPU lacks. So a wrong mapping fails on every machine missing the feature, which is most of them.

The precondition you're pointing at is real — the binary must run where it was built — and it now says so in the test. That holds for a test suite (same job compiles and executes) and for the benchmark builds this key describes. A cross-built binary carried to a weaker machine *should* trip it: its `build_isa` would then be naming instructions that machine cannot run, which is exactly the claim the key must not make.

GCC 167/167, Clang 167/167.

PR #447, opened 2026-08-15
fix(bench): stop preflight failing on the harness's own output

Ravenwater:
## The bug

Reported from a real Zen 4 session. The PowerShell driver **builds before it measures**, writes its cmake logs into `benchmarks\data\<machine>\logs`, and preflight then failed the run as a **dirty tree** — tripped by output the harness had just produced itself. The suggested escape hatch (`set ALLOW_DIRTY=1`) doesn't work in PowerShell either, so the run could not be started at all.

A gate a normal run trips by existing is not a gate, it's a wall.

## Four fixes

**1. The logs are run artifacts, not results.** `benchmarks/data/**/logs/` is gitignored, so a harness cannot make the tree dirty with its own diagnostics.

**2. Dirtiness excludes the output directory.** `preflight` gained `--ignore-path` / `-IgnorePath`, and every harness passes its own `OUTDIR`. A previous run's CSVs say *nothing* about whether the source that built the binary is reproducible — counting them stamped `tree_git_dirty=1` on a sidecar whose code was pristine. That's the exact failure this feature exists to prevent, and it would have hit anyone running twice before committing.

This needed `--untracked-files=ALL`: git collapses a wholly untracked directory into one entry (`?? benchmarks/`), which no prefix filter for `benchmarks/data/<machine>/` can match. The first version of the filter **silently did nothing** for that reason — the test below is what caught it.

**3. `set ALLOW_DIRTY=1` is cmd.exe syntax.** In PowerShell `set` aliases `Set-Variable`, so it creates a shell variable and the run fails again with the same message and no clue why. Every PowerShell harness now takes `-AllowDirty`, and the hint names `$env:ALLOW_DIRTY = "1"` and says plainly that the cmd form does nothing. Both shells now also **list what is dirty** — without it you have to go looking, and the answer is usually a file a harness or an editor dropped.

**4. `run_blocking_ab.ps1` gates before the build.** It was the only harness gating afterwards, which is both slower to fail and the reason it saw its own logs. It now gates first (seconds, not a full configure-and-build) and again after, since the builds heat the machine and the temperature that belongs in the sidecar is the one the measurements start at.

## Test Results

Verified against an isolated repo, in all three directions that matter:

| Scenario | Expected | Result |
|---|---|---|
| Previous run's CSVs present, source clean | pass, `tree_git_dirty=0` | rc=0, `tree_dirty_excluded=benchmarks/data/m1` |
| A real source edit | fail | rc=1, lists ` M src.c` |
| Untracked file outside `OUTDIR` | fail | rc=1, lists `?? stray.txt` |

The `-uall` fix is why row 1 passes: without it the filter matched nothing and row 1 reported dirty.

| Suite | Result |
|---|---|
| GCC | 167/167 |
| Clang | 167/167 |

`bench_powershell_parse` covers the four `.ps1` files on every CI platform that has a PowerShell.

## Test plan

- [ ] Tier 1 CI on 8 platforms
- [ ] Zen 4 re-run gets past preflight without `-AllowDirty`

Relates to #442

## Summary by CodeRabbit

- **New Features**
  - Added options to allow benchmark runs with uncommitted changes and exclude generated output paths from checks.
  - Added configurable parallel job counts for sweep runs.
  - Improved preflight validation with clearer reporting and remediation guidance.

- **Documentation**
  - Updated benchmark guidance for dirty-tree overrides, ignored output directories, and PowerShell environment-variable syntax.

- **Chores**
  - Excluded generated benchmark logs from version-control tracking.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,13 @@ examples/phase15_sparse_direct/data/
 benchmarks/data/klu/
 benchmarks/data/superlu/
 
+# Configure/build logs the PowerShell harnesses write beside their CSVs. These
+# are run artifacts, not results -- and while they were tracked, a harness that
+# builds before it measures made the tree dirty WITH ITS OWN OUTPUT, so preflight
+# then refused to measure. A gate that a normal run trips by existing is not a
+# gate, it is a wall.
+benchmarks/data/**/logs/
+
 # IDE
 .idea/
 .vscode/
--- a/benchmarks/preflight.ps1
+++ b/benchmarks/preflight.ps1
@@ -47,7 +47,23 @@ param(
     [ValidateSet('before', 'after')][string]$Phase = 'before',
     [int]$Threads = 0,
     [string]$Repo = '',
-    [switch]$ReportOnly
+    [switch]$ReportOnly,
+    # Permit a dirty working tree (still recorded as tree_git_dirty=1).
+    #
+    # A SWITCH, not just the ALLOW_DIRTY environment variable, because the
+    # obvious way to set that variable does not work here: `set ALLOW_DIRTY=1` is
+    # cmd.exe syntax, and in PowerShell `set` is an alias for Set-Variable, so it
+    # creates a shell variable (or errors) and the run fails again with the same
+    # message. The environment form is $env:ALLOW_DIRTY = "1"; this switch means
+    # nobody has to know that.
+    [switch]$AllowDirty,
+
+    # Exclude a directory from the dirty check -- a harness passes its own
+    # OutDir. A previous run's CSVs, and the cmake logs these harnesses write
+    # beside them, say NOTHING about whether the source that built the binary is
+    # reproducible; counting them would stamp tree_git_dirty=1 on a sidecar whose
+    # code was pristine, and a field that lies is worse than one that is missing.
+    [string]$IgnorePath = ''
 )
 
 $ErrorActionPreference = 'Continue'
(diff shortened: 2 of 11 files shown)


Ravenwater:
Correct, and fixed in `22e3f38`.

The exclusion took **any** directory, so pointing it at a source tree removed edits there from the dirty check. `--ignore-path benchmarks` would have let a modified `preflight.sh` — the gate itself — be recorded as `tree_git_dirty=0`. That is a hole straight through the gate, and precisely the "recorded fact that is not true" failure this whole feature exists to prevent. It was a weaker exclusion than my own comment claimed.

Now contained to `benchmarks/data/<something>` — a directory that cannot hold source. Two details of the fix worth stating, because both could have gone the other way:

- **In-repo paths outside that root are refused, not silently narrowed.** Quietly ignoring the request would leave a run measured under a gate the operator thinks is stricter than it is. It warns and keeps the full dirty check, so an unexpected `OUTDIR` is *blocked* rather than quietly measured.
- **Paths outside the repository need no handling at all** — git never reports them, so there is nothing to exclude and no warning to give.

Verified on an isolated repo, including the attack you described:

```
legitimate outdir, run output only        -> tree_dirty_excluded=benchmarks/data/m1, dirty=0, rc=0
--ignore-path benchmarks + a source edit  -> refused with a warning, dirty=1, rc=1
same edit, legitimate outdir              -> still caught, rc=1  (" M benchmarks/preflight_like_source.sh")
outdir outside the repo                   -> no exclusion, no spurious warning
```

The contract test now asserts it on **every** CI platform, in both implementations: asking to exclude `benchmarks` must not emit `tree_dirty_excluded`. Verified negatively — widening the pattern back to any in-repo directory fails the test:

```
[pattern widened to "$repo"/*/ ]  -> ***Failed
[restored]                        -> Passed
```

GCC 167/167, Clang 167/167.

PR #452, opened 2026-08-16
feat(bench): integrity guards, and the cache-blocking study written up

Ravenwater:
## Summary

The four-arm run refuted the hypothesis it was built to test, and found two defects in the harness on the way. Both are now guarded, and the study is written up.

## Guard 1 — `ROUNDS` must be a multiple of the arm count

The rotation only cancels position effects when every arm leads equally often. At **5 rounds over 4 arms** the first arm leads twice and one never leads — a residual bias of exactly the size the rotation exists to remove. Both harnesses round up and say so; the sidecar records what actually ran.

## Guard 2 — identical arms that disagree fail the analyzer

Arms collapse onto each other on particular hardware: on Zen 4 `kconly` **is** `default` (32 KB L1d matches the model), on the Jetson `detected` **is** `kconly`. Those pairs are null comparisons between identical binaries — whatever they measure is the session, not the code, and it is the floor under everything else that session can claim.

Judged on the **median** deviation, not point by point: individual points scatter a few percent on any machine, and gating on the worst would fail every healthy session.

```
zen4   detected vs mconly   median 1.2%  → passes
jetson detected vs kconly   median 3.9%  → INTEGRITY FAILURE (5 points past 9%)
```

The runner now prints the identical-arm pairs it finds, so the check is discoverable rather than something you have to think of.

## The study

`docs/performance/cache-blocking-ab-study.md` — design, hypothesis, exact sign and sign-flip permutation tests on paired log-ratios, identical-binary calibration, verdict, mechanism, next experiment.

**H₁(kc) — "the loss is kc, mc is harmless" — is refuted and reversed.**

| i7, T=1 | kc | mc | median | same-sign | p |
|---|---|---|---|---|---|
| `kconly` | 256→**384** | 64→42 | **0.999** | 3/5 | 0.69 |
| `mconly` | 256 | 64→**320** | **0.919** | 5/5 | 0.0625 |

Paired directly, `kconly` beats `mconly` by 7.4% (n=10, **p = 0.027**). The first round got it backwards by inferring from the arm where *both* move, then generalising one machine's behaviour into a rule.

Two findings worth keeping:

- **`mc_used` (#448) shows the effect tracking the parameter, not the shape.** Within *one* arm: the points where the partition clamped mc *below* default lost nothing (1.001, 0.987), while every point where it grew lost, monotonically to 0.799.
- **The Jetson session is disqualified by its own data** — 12.8% median round-to-round spread at 15 W with `schedutil`, larger than the effect under test.

Also documented: n=5 cannot produce p<0.05 in a two-sided exact test (the floor is 2/2⁵ = 0.0625), so per-regime results at that value are at the design's resolution limit, not weak effects.

## Docs structure

**Performance studies lifted out of Design to a top-level section** — they answer "what did the machine do", not "how is it built", and are cited from issues often enough to deserve a top-level home. The three existing pages moved to `docs/performance/` with every inbound link repointed. Site builds clean (61 pages, verified locally with a real `npm run build`).

`systems.md` now records the correction rather than the refuted claim.

## Test Results

| Suite | Result |
|---|---|
| GCC | 170/170 |
| Clang | 170/170 |
| Analyzer gate | fires on the Jetson pair, passes 3 healthy identical pairs and all normal comparisons |
| Docs site | builds, 61 pages |

Relates to #430

## Summary by CodeRabbit

* **New Features**
  * Added performance case studies covering cache blocking, multicore scaling, and threading benchmark results.
  * Added consistency checks and improved validation for cache-blocking benchmark runs.
  * Benchmark scripts now balance rotated rounds automatically and report any adjustments.

* **Documentation**
  * Reorganized performance studies into a dedicated documentation section.
  * Updated navigation and references to reflect the new documentation paths.
  * Documented corrected cache-blocking findings and multicore scaling results.


Ravenwater:
Correct, and fixed in `72b3d9b`.

The CSV has grown columns twice — `pool`, then the `mc_used` group — so `kc` sits at field 9 in the pre-`pool` files still committed under `benchmarks/data` and at field 10 in current ones. My positional read returns the wrong pair on the older layout, verified against both:

```
jetson-orin (old header)   positional -> 16,2048     by name -> 1024,16
i7-12700k   (new header)   positional -> 384,213     by name -> 384,213
```

Worth being precise about the impact, because it argues *for* the fix rather than against it: today the runner only ever reads CSVs it just wrote, so both arms shift together and the grouping still comes out right. It is correct by accident, and the accident is one column away from ending. The analyzer already parsed by name, so the two halves of the same consistency check disagreed on method — which is the part that would have made a future failure hard to see.

Both runners now index the header.

**Also renamed the PowerShell helpers to approved verbs** — `Get-ArmExe`, `Get-ArmCsv`, `Get-ArmBlocking`. PSScriptAnalyzer flagged only the new function, but `Arm-Exe` and `Arm-Csv` are the same finding, and mixed naming inside one file is worse than either alone.

Verified end to end: the consistency-check discovery still finds the identical-parameter pair on a fresh two-arm session.

GCC 170/170, Clang 170/170.

Issue #453, opened 2026-08-16
perf(simd): mc ignores the C strip -- the L2 model counts A only


Ravenwater:
## Problem

`derive_blocking` sizes `mc` so that the packed A block fills about half of L2:

```
mc ~ l2_bytes / (2 * kc * sizeof(T))
```

That counts **A and nothing else**. The same cache simultaneously holds the strip of C being accumulated, which is `mc × nc` — and on real parameters C dominates A by an order of magnitude:

| i7-12700K, per ic block | A block `mc·kc` | C strip `mc·nc` | L2 |
|---|---|---|---|
| shipped default, mc=60 | 123 KB | 1.97 MB | 1.25 MB |
| detected L2, mc=318 | 651 KB | **10.4 MB** | 1.25 MB |

So the term the model optimises is not the term that fills the cache.

## Evidence

From the four-arm A/B ([study page](https://github.com/stillwater-sc/mtl5/blob/main/docs/performance/cache-blocking-ab-study.md), #430): **raising `mc` is the single largest harm, and it is visible with one thread.**

| i7, T=1 | kc | mc | median ratio | same-sign | p (exact perm) |
|---|---|---|---|---|---|
| `kconly` | 256→384 | 64→42 | 0.999 | 3/5 | 0.69 |
| `mconly` | 256 | 64→**320** | **0.919** | 5/5 | 0.0625 |

And within the *single* `mconly` arm, the effect tracks the mc the loops actually stepped by (`mc_used`, #448) rather than the shape:

| shape, T | mc used vs default | ratio |
|---|---|---|
| 256×8192, T=8 | **32** (½×) | 1.001 |
| 1024³, T=8 | 128 (2×) | 0.799 |
| 2048²/4096², T=8 | 256 (4×) | 0.856 / 0.901 |
| all shapes, T=1 | 318 (5×) | 0.862 – 0.970 |

Monotone in mc, in one binary, in one session. A model whose `mc` term is wrong in exactly this direction is the natural explanation.

## Proposed change

Bound `mc` by the working set the ic loop actually touches:

```
mc * (kc + nc) * sizeof(T)  <=  fraction * l2_bytes
```

instead of `mc * kc * sizeof(T) <= fraction * l2_bytes`.

Consequences to check rather than assume:
[...]

Issue #451, opened 2026-08-16
epic(simd): integer lanes for dot / gemv / gemm (int8, int16, int32)


Ravenwater:
## Problem

MTL5's dense path is floating-point only, by construction:

```cpp
// include/mtl/simd/batch.hpp
static_assert(std::is_floating_point_v<T>,
              "mtl::simd::batch currently supports floating-point lanes ...
               integer lanes are a future extension.");
```

Every SIMD kernel — `dot`, `gemv`, the GEMM micro-kernel — is therefore unavailable for integer element types. An integer matrix falls back to the scalar path, i.e. whatever the autovectorizer manages.

That leaves the **largest remaining performance lever in the dense path** unused. Lane count alone:

| type | AVX-512 lanes | dot-product ISA |
|---|---|---|
| fp64 | 8 | — |
| fp32 | 16 | — |
| int16 | 32 | `vpdpwssd` |
| int8 | 64 | `vpdpbusd` (VNNI), 4 MACs/lane |

Zen 4 (`ryzen-9-8945hs`) has AVX512-VNNI; the Jetson's Cortex-A78AE has SDOT/UDOT. The ceiling over fp32 is roughly 4× from lanes and up to 4× again from dot-product instructions — an order of magnitude more than the ±5% the cache-blocking work (#430) is chasing.

It also fits what this library is for: *"designed for mixed-precision algorithm design with custom number types"*.

## Scope

**In:** `dot`, `gemv`, `gemm` — the BLAS operations that need a **ring**, not a field.

**Out:** every factorization. LU/QR/Cholesky/LDLᵀ/SVD/eigen/inv divide, and integer division truncates. #450 makes that structural — `Field` now excludes the integers and the 25 dividing entry points require `FieldMatrix`, so the unstable case no longer compiles rather than being avoided by convention.

## The design decisions, before any code

**1. Accumulator width is arithmetic, not preference.**

| operands | accumulator | safe depth | verdict |
|---|---|---|---|
| int8 × int8 | int32 | ~65 000 terms | **the valuable case** — what VNNI/SDOT implement |
| int16 × int16 | int32 | ~2 terms | needs int64; safe past 1M there |
| int32 × int32 | int64 | depends on RANGE, not on the type | fine for realistic data |

[...]

PR #469, opened 2026-08-22
feat(simd): the VNNI micro-kernel, and the claim it falsifies

Ravenwater:
`vpdpbusd` consumes **four k-values per instruction**, which needs a quad-interleaved pack layout and a micro-kernel whose left operand is a *broadcast quad* rather than a broadcast scalar. #468 built neither and said so. This builds both.

## The result revises #468's headline

#468 concluded: *"on a machine without VNNI there is no int8 GEMM win at all."* That was true of the **widen-on-load** kernel — the only int8 GEMM that existed when it was written — and it does not survive changing the kernel.

Same Xeon E5-2420 v2, SSE4, n = 512, the instruction still **decomposed**:

| arm | GOP/s | vs widen |
|---|---|---|
| `gemm_f32` | 19.5 | 1.47× |
| `gemm_i32` | 12.2–14.0 | — |
| `gemm_i8_i32` (widen-on-load) | 13.2 | 1.00× |
| `gemm_i8_i32_quad` | **16.6** | **1.26×** |
| `gemm_u8i8_i32_quad` | **20.7–21.7** | **~1.6×** |

The native `u8 × i8` shape beats fp32 — on a 2013 part with no VNNI silicon in it.

The mechanism was **checked in the disassembly, not inferred**: Highway's decomposition is a pair of `vpmaddwd` plus sign-extension shifts, which still folds four products into each accumulator lane in a handful of instructions, where widen-on-load runs four independent promote-multiply-add chains. The symmetric form carries more of that shift work than the mixed one (4 `vpsraw` + 2 `vpsllw` against 2 + 1 plus a mask) — exactly the 16.6-against-21.7 gap.

**What survives:** the operand *width* still contributes nothing to a GEMM. **What was wrong:** equating "the instruction" with "VNNI silicon". Most of the win here is the four-products-per-lane *kernel shape*, which a machine without the instruction can partly express anyway. So a VNNI part now has a narrower question to answer — what the native instruction adds *on top of* the quad kernel, against that same kernel decomposed on the same machine — and `hardware-expansion-plan.md` is corrected to ask it that way rather than to expect the whole 1.6×.

## The selection is explicit, and that is the interesting bug

Kernel selection was first *inferred from the element types*. Since `(i8,i8)` is valid input to **both** kernels, that silently rerouted `mult`'s widening path — through a completely green test suite, because the two kernels agree **bit for bit** and no result can tell them apart.

The benchmark caught it: the two int8 arms started reporting the same number to three digits. They had become the same kernel, and the arm meant to be the control had stopped being one.

Now: `gemm_blocked` takes a `gemm_kernel` template argument defaulting to `widening`, `mult` is unchanged, the quad kernel is reached only through the new `mult_quad`, and a **structural** test pins the default — since a result-based one provably cannot. Both arms compile into one binary on purpose: comparing a quad number on one machine against a widen number on another is the same missing-control error that cost the dot headline 20% once already.

## What's in it

- **`batch::quad_dot_broadcast_accumulate`** — a GEMM needs the same four A-values against a whole row of B, so the left operand is a 32-bit splat of four bytes. Its lane assignment is checked **exactly, per lane**; the dot form's weaker *"the total is right"* contract would not have caught a transposed quad, because for a GEMM a lane **is** a column of C — a permutation there is a wrong answer, not a reordering.
- **`detail/gemm_quad_pack.hpp`** — `Ap[…][i*4+q] == A(i, 4g+q)`, k zero-padded to a multiple of four. Keeps generic strides, edge padding, and independent panels for cooperative threaded packing.
- **`detail/gemm_quad_microkernel.hpp`** — alpha applied to the int32 tile on the way out, *not* folded into the packed panel (folding truncates it into a byte).
- **`gemm_blocked<TC, TA, TB, Kern>`** — one nest, three kernels; `TA`/`TB` split because `(u8,i8)` is VNNI's native shape. All the threading, barriers, grid planning and exception handling are shared.
- **`mtl::mult_quad`** + two new benchmark arms.

`(i8, u8)` is **rejected rather than reordered**: a dot product is symmetric so swapping fixes it there, but a GEMM is not, so there is no swap to suggest. This also means column-major C — computed as `C^T = B^T A^T` — cannot take the asymmetric pairing through the kernel and falls to the generic loop. Tested.

## Known lever, deliberately untouched

`kc` is still derived for a TC-wide B micro-panel while the quad panel holds bytes, so it could be several times larger. Left alone and documented: #430 and #453 both found that moving a cache-derived parameter moves *block counts* the thread partition is sensitive to, so this gets its own measurement rather than riding along with a new kernel.

## Testing

179/179 (was 176) on **gcc+Highway**, **clang+Highway** and the **scalar fallback**, plus **UBSan** over the integer kernels on both backends.

⚠️ The benchmark box is noisy — figures above are best-of-8 interleaved rounds (the two quad arms were the *most* stable of the six). A committed number should come from `run_int_bench.sh` on a quiet machine.

Relates to #451

## Summary by CodeRabbit

- **New Features**
  - Added quad-optimized integer GEMM support for signed 8-bit and mixed unsigned/signed 8-bit operands.
  - Added a dedicated `mult_quad` operation with support for common matrix layouts, scaling, threading, and accumulator behavior.
  - Added optimized packing and SIMD execution for supported workloads, with safe fallback behavior when conditions are not met.

- **Performance**
  - Improved int8 GEMM throughput on supported hardware, including systems without specialized VNNI instructions.

- **Documentation**
  - Updated benchmark and performance guidance with results, limitations, and validation details.

### DIFF
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -282,15 +282,57 @@ memory saving it was built for is irrelevant once operands are reused. Integer
 multiply is also simply slower than float FMA on this ISA — there is no integer
 FMA, so it is a separate multiply and add.
 
-The consequence is worth stating plainly, because it decides what hardware is
-worth buying: **on a machine without VNNI there is no int8 GEMM win at all.**
-Everything an int8 GEMM can offer over fp32 has to come from the instruction —
-which makes a VNNI or AMX machine's measurement almost purely a measurement of
-the instruction, cleanly separated from bandwidth. That is the opposite of the
-dot suite, where the instruction was ~1.2× against ~18× of traffic.
-
 `gemm_i8_i32` is the **widen-on-load** path: narrow operands promoted to int32
 lanes, then an ordinary multiply-add. It is *not* `vpdpbusd`, which consumes
 four k-values per instruction and needs a quad-interleaved pack layout and a
-different micro-kernel. That remains unbuilt, and this arm is what it would be
-measured against.
+different micro-kernel.
+
+### The quad arms, and what they revise
+
+That micro-kernel now exists, and it changes the conclusion above. The two new
+arms run the **quad multiply-accumulate** itself — four k-values per
+instruction, from quad-interleaved panels — against the same operands:
+
+| arm | GOP/s | vs `gemm_i8_i32` |
+|---|---|---|
+| `gemm_f32` | 19.7 | 1.49× |
+| `gemm_i32` | 14.0 | 1.06× |
+| `gemm_i16_i32` | 12.8 | 0.97× |
+| `gemm_i8_i32` (widen-on-load) | 13.2 | 1.00× |
+| `gemm_i8_i32_quad` | **16.6** | **1.26×** |
+| `gemm_u8i8_i32_quad` | **21.7** | **1.64×** |
+
+Same Xeon E5-2420 v2, SSE4, n=512, `int8 quad dot: decomposed` — **this machine
+still has no VNNI.** Best of 8 interleaved rounds; the box is noisy and
+individual arms occasionally lose a scheduling slot, though the two quad arms
+were the *most* stable of the six. A committed figure should come from
+`run_int_bench.sh` on a quiet machine.
+
+**The claim this replaces** was that on a machine without VNNI there is no int8
+GEMM win at all. That was true of the *widen-on-load* kernel and does not
+survive changing the kernel: the quad path beats it by 1.26× symmetric and
+1.64× in VNNI's native `u8 × i8` shape — the latter also beating fp32, on a
+2013 part with no VNNI silicon anywhere in it.
+
+The reason is visible in the disassembly rather than inferred. Highway's
+*decomposition* of the quad accumulate is a pair of `vpmaddwd` plus
+sign-extension shifts, which still folds four products per accumulator lane in a
+handful of instructions — where widen-on-load runs four independent
+promote-multiply-add chains. And the symmetric `i8 × i8` form carries visibly
+more of that shift work than `u8 × i8` (4 `vpsraw` + 2 `vpsllw` against 2 + 1
+plus a mask), which is exactly the 16.6-against-21.7 gap.
+
+So the operand width still contributes nothing, and the earlier reading of that
+— "everything an int8 GEMM offers must come from the instruction" — stands. What
+was wrong was equating *the instruction* with *having VNNI silicon*: most of the
+win here is the four-products-per-lane **kernel shape**, which a machine without
+the instruction can still partly express. A VNNI or AMX machine's GEMM number
+therefore remains an almost pure measurement of arithmetic rather than
+bandwidth, but its baseline is now `gemm_i8_i32_quad` on the same machine, not
+the widening arm.
+
+Both int8 arms are compiled into the same binary on purpose. They compute
+bit-identical results — integer addition is associative — so nothing in an
+*answer* distinguishes them, and comparing a quad number on one machine against
+a widen number on another is precisely the missing-control error that cost this
+programme 20% on the dot headline once already.
--- a/benchmarks/harness/runner.hpp
+++ b/benchmarks/harness/runner.hpp
@@ -61,9 +61,10 @@ namespace mtl::bench {
 // the traffic, and the SHAPE of the curve across them is the actual result.
 //
 // The honest place for VNNI's arithmetic density is a GEMM, where operands are
-// reused O(n) times and the kernel is compute-bound throughout. MTL5 has no
-// integer GEMM -- phases 0-3 delivered dot and gemv -- so there is no int gemm
-// arm here rather than a misleading one. See bench_gemm_int below.
+// reused O(n) times and the kernel is compute-bound throughout. That is
+// bench_gemm_int below, which since phase 5 runs the quad multiply-accumulate
+// itself -- not merely narrow operands widened on load -- against the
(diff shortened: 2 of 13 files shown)


Ravenwater:
All nine review findings verified against the code and fixed in 1ca3ee8. Every one was real; three were more than cosmetic.

### Undefined behaviour in the quad pack (`gemm_quad_pack.hpp`)

Correct, and I'd underline how it was found. Both routines hoisted a source pointer out of the padding guard:

```cpp
const T* row = A + (i0 + i) * rs;             // formed for i >= mr too
... = (i < mr && p + t < k) ? row[...] : 0;
```

The value is never loaded, but *forming* the pointer is already undefined once the row is past the end of the source. The whole index now sits inside the guard, in both `pack_A_quad` and `pack_B_quad_panels`.

Worth noting that **neither sanitizer would have caught this**: ASan sees no bad load because there isn't one, and UBSan's `pointer-overflow` check tests wraparound rather than escape from the object. My UBSan runs were clean before the fix and after it. Found by reading.

### `mult_quad` promised a fallback it could not reach

This one was a documented guarantee that was false. The doc comment says the non-dense case *"falls back to the accumulator-aware generic loop, so this is always correct, never merely fast."* It wasn't: `is_row_major_v<MC>` was named at function scope, outside any discarded branch, and `Matrix` does not require `orientation` — `compressed2D` and `transposed_view` both lack it. So the fallback path was a **hard compile error for exactly the types it existed to serve**.

The layout gate now tests for `orientation` before using it. I checked the new regression test actually has teeth: **9 errors against the previous header, 0 against this one.**

### The 1.64× was a two-variable ratio

Accepted, and this is the one I'm most glad you caught — the PR reproduces the exact error its own narrative condemns. `gemm_u8i8_i32_quad ÷ gemm_i8_i32` moves the kernel **and** the operand signedness.

| ratio | what varies |
|---|---|
| **1.26×** | the kernel, operands fixed (`i8×i8` quad ÷ `i8×i8` widen) |
| **1.31×** | operand signedness, kernel fixed (`u8×i8` quad ÷ `i8×i8` quad) |
| 1.64× | **both — not a result** |

§6's own headline correction (1.42× → 1.17×) came from this same mistake one kernel earlier. All three documents now lead with 1.26× as the kernel result and label the 1.64× as what it is: *"best available after, against best available before"* — which exists only because `load_widen` requires matching signedness, so there is no `u8 × i8` widening arm to compare with. The preceding instruction-only conclusion is now explicitly scoped to the widen-on-load kernel.

### The benchmark did not read its own output

Right, and this is the guard that would have caught the defect the previous commit had to find by noticing two rows agreeing to three digits. `verify_gemm_int` now samples C against a 64-bit reference after **every** integer arm (not just the new two — the gap was uniform), outside the timed region. It **fails** the run rather than warning, since a sidecar from a build that computes the wrong answer is worse than none because it looks like data. Verified by tampering with the reference:

```
INTEGRITY FAILURE: gemm_i32 at n=64 -- C(0,0) = 195040, expected 195041
The kernel under test computes the wrong answer; no results written.
```

Sampled rather than exhaustive because a full O(n³) scalar reference at n=1024 costs more than the measurement it guards; the errors that survive to this point are structural.

### I named an instruction the binary does not contain

Correct. The runner claimed the symmetric form is emulated *"with two `vpdpbusd` plus a shift and subtract"* on every x86 below AVX10.2. That describes `HWY_AVX3_DL` only — SSE4 and AVX2 have no `vpdpbusd` at all and decompose through `vpmaddwd`, which **my own disassembly in the previous commit shows**. The comment is now split by target.

### The Zen build-flag claim

Also right, and wrong in the way that matters. `--allow-decomposed` only lets the runner *time* a decomposed build; `-march=znver4` takes the native path. Forcing the decomposed arm means dropping to `-march=x86-64-v3`, which removes AVX-512 **along with** VNNI — so the two arms would differ in vector width and register count as well as the instruction. That is another two-variable comparison. The plan now says an honest within-machine measurement of the instruction alone needs a Highway AVX-VNNI path or an AVX10.2 part, rather than implying Zen 4 settles it, and requires any `znver4`-vs-`v3` run to record `build_isa` and the arm labels and be reported as two-variable.

### Smaller items

- **Mis-named wrap test** — correct, `255 × 127 × 5000` fits `int32` comfortably and the comment contradicted itself. Renamed to the headroom claim it actually checks, and it now asserts `CHECK_FALSE(any_negative)` against the next test's `CHECK(any_negative)`, so the pair brackets the boundary.
- **Alpha across a `kc` split** (nitpick) — valid: `K = 13` is one `kc` block, so the test never exercised the per-block scaling it was written for. Added, with `alpha=3, beta=5` and `K = 2·kc + 7`.
- **`<algorithm>`** — added directly.

Re-verified: **179/179** on gcc+Highway, clang+Highway and the scalar fallback, plus UBSan over the integer kernels on both backends.

Ravenwater:
Both findings on `802cd47` / `94f6ced` fixed in 92d1f47.

**Stale n=128 ratio in the Xeon profile** — correct. The header still quoted 1.10× from the hand-typed run that the profile run replaced; the committed CSV says 1.19×. The three performance documents were updated when the data was and this comment was missed, which is precisely the drift the committed profile exists to prevent. Checked for other instances: none.

**Bare code fence (MD040)** — correct, it is console output, so `text`. The remaining bare fences in that README are closing delimiters.

Branch is green: 42 checks pass, 2 skipped. Ready to merge once your review of the last commits lands.

